PR #53, opened 2026-04-05
fix: keep completed notification visible for 30s grace period

Octane0411:
## Summary
- Completed sessions were being removed by `removeInvisibleSessions()` during process monitoring cycles (every few seconds), causing the completion notification card to vanish before the 10s auto-collapse timer could fire
- Added a 30s grace period in `isVisibleInIsland` so completed sessions remain visible long enough for the notification to display and auto-collapse naturally

## Test plan
- [x] `swift build` passes
- [x] All 118 tests pass
- [ ] Manual: trigger a session completion and verify the notification stays visible for ~10s before auto-collapsing

### DIFF
--- a/Sources/OpenIslandCore/AgentSession.swift
+++ b/Sources/OpenIslandCore/AgentSession.swift
@@ -375,12 +375,20 @@ public extension AgentSession {
         attachmentState.isLive
     }
 
+    /// Grace period after completion before the session becomes invisible.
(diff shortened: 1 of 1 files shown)


Octane0411:
Closing: the 30s grace period was a workaround, not a root-cause fix. Replacing with proper process liveness tolerance.

PR #90, opened 2026-04-06
feat: add Dock icon toggle in General > 行为 settings

seedoilz:
feat: add Dock icon toggle in General > 行为 settings

Users can decide whether show the icon of the app in the dock or not when the app is running in the background.

### DIFF
--- a/Sources/OpenIslandApp/AppModel.swift
+++ b/Sources/OpenIslandApp/AppModel.swift
@@ -8,6 +8,7 @@ import SwiftUI
 @Observable
 final class AppModel {
     private static let soundMutedDefaultsKey = "overlay.sound.muted"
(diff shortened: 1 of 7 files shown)


Octane0411 (commented):
Thanks for the PR! Could you describe why you changed the main.swift files (removing @main)? The rest of the changes look good to me.

Octane0411 (commented):
Thanks for the PR! Could you describe why you changed the main.swift files (removing @main)? The rest of the changes look good to me.

seedoilz:
Actually, I can not build the project correctly when using `@main`. So i have changed a little. I will reset this change and adapt the i18n feature.

Octane0411:
Could you change the default value of `showDockIcon` to `true`? We'd like the app to show in the Dock by default.

PR #247, opened 2026-04-10
feat: add tmux session/window/pane jump support

### LINKED ISSUE #235: [Feature Request] ghostty 中使用 tmux, 在 tmux 中打开 claude-code, open-vibe-island 双击会跳转到 iterm2
如果是 ghostty 新开一个 tab 打开 claude-code, 则双击能跳转到对应的 cc 所在的 ghostty tab.

Expect
- 希望能跳转到 ghostty 中的 tmux 对应的 pane
- 如果不可以的话，希望在 ghostty + tmux 的情况下能禁用双击跳转功能（如果能检测到的话）

graelo:
Hi, thanks for this project! Super happy to see this open source alternative ❤️ 

I did see the CLAUDE.md file mention that tmux was out of scope, but I took
it as an instruction for agents, not a project-level restriction. As I exclusively use 
tmux, I went ahead. I was further motivated by the author of issue #235.

I hope this is the right time.

## Summary

When an agent (Claude Code, OpenCode, etc.) runs inside a tmux pane, clicking
its card in the Island UI now navigates to the exact tmux session, window, and
pane where the agent is running. Previously, clicking only activated the
terminal app (e.g. Ghostty) without switching the tmux context.

This required plumbing tmux pane information through the full discovery →
session → jump pipeline, and handling the constraint that GUI apps have no
`$TMUX` environment — so all tmux commands must explicitly target a client.

## Changes

- Add `tmuxTarget` and `tmuxSocketPath` fields to `JumpTarget` and
  `ProcessSnapshot` so tmux pane info flows end-to-end through the pipeline.
- Wire tmux fields from `ProcessSnapshot` into `JumpTarget` in
  `ProcessMonitoringCoordinator.syntheticClaudeSession()` for Claude Code
  sessions discovered via process monitoring.
- Broaden the `TerminalJumpTargetResolver` tmux candidate filter to include
  sessions with only a `terminalTTY` (no existing `tmuxTarget`), enabling
  tmux pane discovery for OpenCode/BridgeServer sessions via TTY matching.
- Add tmux pane fetching to `TerminalJumpTargetResolver` using `list-panes -a`
  with `session:window.pane` target format and a printable multi-char
  separator (`<|>`) — tmux converts control characters like `\t` and `\x1f`
  to printable equivalents when passed as raw bytes via Process arguments.
- Add `jumpToTmuxPane()` to `TerminalJumpService` executing the sequence
  `switch-client` → `select-window` → `select-pane`, with explicit client
  targeting via `list-clients` (required from GUI apps without `$TMUX`).
- Add `resolveTmuxInfo()` in `ActiveAgentProcessDiscovery` to discover tmux
  target, socket path, and host terminal app from the agent's TTY.

## Tested workflows

- Jump from a different window within the same tmux session → switches to
  the correct window and pane.
- Jump from a different tmux session entirely → switches session, window,
  and pane.
- Jump when Ghostty is on another macOS desktop space → brings Ghostty to
  focus and switches to the correct tmux context.
- Multiple agent panes in the same window (e.g. two OpenCode instances) →
  selects the correct pane matching the clicked agent card.

## Additional note

I'm not a huge fan of the mutating `ProcessSnapshot`, but preparing all the fields
upfront was not very nice either. Given the pace of the project, I expect all of this
will be refactored soon anyway. In my view, having a fully working tmux jump feature
is both a differentiation factor and a guard against drifting away from being able to 
support it in the future.

Closes #235

### DIFF
--- a/Sources/OpenIslandApp/TerminalJumpTargetResolver.swift
+++ b/Sources/OpenIslandApp/TerminalJumpTargetResolver.swift
@@ -29,6 +29,12 @@ struct TerminalJumpTargetResolver {
         var ttyName: String?
     }
 
+    struct TmuxPaneSnapshot: Sendable {
+        var paneID: String
+        var tty: String
+        var title: String
+    }
+
     private static let appleScriptTimeout: TimeInterval = 3
     private static let fieldSeparator = "\u{1F}"
     private static let recordSeparator = "\u{1E}"
@@ -49,8 +55,29 @@ struct TerminalJumpTargetResolver {
             let name = normalizedTerminalName(for: $0.jumpTarget?.terminalApp)
             return name == "kaku" || name == "wezterm"
         }
+        // Tmux candidates: sessions that already have a tmuxTarget, OR sessions
+        // whose terminalTTY maps to a tmux pane (e.g. OpenCode sessions created
+        // by BridgeServer without tmux info). We discover the mapping below.
+        let tmuxSessions = sessions.filter {
+            $0.jumpTarget?.tmuxTarget != nil || $0.jumpTarget?.terminalTTY != nil
+        }
 
         var jumpTargetUpdates: [String: JumpTarget] = [:]
+
+        // Tmux: match sessions and resolve their tmux pane info.
+        // Also discovers tmux targets for sessions that only have a TTY.
+        if !tmuxSessions.isEmpty {
+            let tmuxSnapshots = fetchTmuxSnapshots()
+            if let snapshots = tmuxSnapshots {
+                let matched = matchTmuxSnapshots(snapshots, to: tmuxSessions)
+                for (sessionID, snapshot) in matched {
+                    if let session = sessions.first(where: { $0.id == sessionID }),
+                       let corrected = correctedTmuxJumpTarget(for: session, snapshot: snapshot) {
+                        jumpTargetUpdates[sessionID] = corrected
+                    }
+                }
+            }
+        }
 
         // Ghostty: match sessions to AppleScript snapshots.
         if !ghosttySessions.isEmpty || sessions.contains(where: { needsGhosttyProbe($0) }) {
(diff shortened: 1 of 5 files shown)


Sources/OpenIslandApp/TerminalJumpTargetResolver.swift
```diff
@@ -49,9 +55,30 @@ struct TerminalJumpTargetResolver {
             let name = normalizedTerminalName(for: $0.jumpTarget?.terminalApp)
             return name == "kaku" || name == "wezterm"
         }
+        // Tmux candidates: sessions that already have a tmuxTarget, OR sessions
+        // whose terminalTTY maps to a tmux pane (e.g. OpenCode sessions created
+        // by BridgeServer without tmux info). We discover the mapping below.
+        let tmuxSessions = sessions.filter {
+            $0.jumpTarget?.tmuxTarget != nil || $0.jumpTarget?.terminalTTY != nil
```
Octane0411, line 62: is this condition too broad?  it matches nearly all sessions, or it should be &&? 
this is checked by claude code, other parts LGTM

graelo: Cool, thanks for digging into this so quickly. I'll reply within the hour.

graelo: Actually, the `||` is intentional. The resolver needs to handle two cases:

- Sessions that already have a `tmuxTarget` (to keep it updated)
- Sessions that only have a `terminalTTY` but no `tmuxTarget` yet: these are the OpenCode sessions created by `BridgeServer` without tmux info. The resolver discovers their tmux pane by matching their TTY against `tmux list-panes` output.

Using `&&` would exclude the 2nd case, which is the main scenario this PR fixes.

In practice, `terminalTTY != nil` is true for most sessions, so the filter is broad, I agree! But the actual tmux matching inside (`matchTmuxSnapshots`) only produces updates for sessions whose `TTY` matches a real tmux pane. Sessions not running in tmux simply get zero matches and are unaffected.

I hope this makes sense!

graelo: Locally, I tried this with claude and opencode sessions in different tmux windows, sessions, etc. Locally, I have 120+ tmux panes across 15 sessions: island with this update keeps working like a breeze.

PS: Why so many sessions: I keep them over time (years!) and back them up and resotre them with a tmux plugin.

Octane0411: merged, Thanks for your contribution!

PR #261, opened 2026-04-11
feat: Warp precision jump — focus the exact tab hosting the agent

loop2zero:
## Summary

Open Island's jump action on a Warp-hosted agent session used to activate Warp.app and stop there — leaving the user to hunt through tabs manually. This PR wires up end-to-end **precision jump**: clicking jump now focuses the exact Warp tab running the target Claude/Codex session.

The scaffolding landed in the earlier commits on this branch (`feat(core): ...` and `feat(app): ...`). This PR adds four fixes discovered and hardened during end-to-end verification, each of which blocked the feature in a different way:

1. **`fix(hooks): use TERM_PROGRAM as authoritative terminal signal`** — Warp sessions were being misidentified as Ghostty when Warp was launched from a Ghostty tab, because `GHOSTTY_RESOURCES_DIR` (and friends) leak across apps through macOS GUI env inheritance. `inferTerminalApp` used to check Ghostty's env var before Warp's, tagging the Warp session as Ghostty, which then ran Ghostty's AX locator and stamped a foreign pane's metadata onto the jumpTarget. `liveAttachmentKey` collided with the real Ghostty owner and the Warp session was deduped out of the island list — **users never saw their Warp sessions**. Fixed by switching detection order to check multiplexers first, then `TERM_PROGRAM` (which cannot leak across apps because each terminal sets it when it execs the shell), then per-app env vars as a last-resort fallback. Regression test pinned.

2. **`fix(core): WarpSQLiteReader matches both firmlink forms of cwd`** — Claude Code captures cwd via `getcwd()` which resolves macOS firmlinks (`/tmp` → `/private/tmp`, `/var` → `/private/var`), while Warp records whatever its shell integration OSCs — usually the unresolved form. The old `WHERE cwd = ?` exact match missed the row whenever an agent ran under `/tmp` or `/var`, dropping `warpPaneUUID` from the jumpTarget and degrading precision jump to bare activation. Now queries both candidate forms in a single statement.

3. **`fix(app): click "Switch to Next Tab" via AX instead of synthesizing Cmd+Shift+]`** — Two synthetic-keystroke implementations failed in different ways: `CGEventPost` silently produced NSBeep (Accessibility/HID filtering drops untrusted synthetic events before keyEquivalent lookup); `NSAppleScript` + `tell application "System Events" to keystroke` cycled exactly one tab on the first invocation after each fresh Accessibility grant, then beeped on every subsequent call (the Apple Event boundary plus the async WindowServer frontmost↔key-window handoff leaves the synthetic event in a state where it reaches Warp's process queue but doesn't walk the normal responder chain). Switched to clicking the menu item directly via AX — bypasses responder chain, key equivalents, modifier state, first-responder identity, and whether a TUI app is running in the focused pane. Verified across three consecutive clean runs.

4. **`fix(app): make jumpToWarpPane robust — wait for frontmost, drop misplaced Accessibility gate`** — (a) Added a short poll loop on `NSWorkspace.frontmostApplication` so `open -b` has time to actually commit Warp as frontmost before the cycle loop starts; previously the first few tab-advance commands raced the activation and cycled nothing. (b) Removed the `accessibilityTrustChecker` gate that used to early-return with "Grant Accessibility permission" — it gated the wrong permission domain and silently killed every dev-binary jump (ad-hoc `swift run` binaries can't practically register for Accessibility). The AX menu click now handles its own permission errors via the existing bounded retry loop and fallback message. (c) Bumped post-click settle delay from 50ms to 100ms for SQLite read stability. (d) Deleted the now-dead `AccessibilityPermissionChecker` wrapper and its unit test.

## Verification

- `swift build` and `swift test` — **158/158 tests green**, including a new regression test for the env-leak misidentification scenario.
- Manual Task 11 verification (from `docs/exec-plans/active/2026-04-10-warp-precision-jump-plan.md`) against a running Warp Stable (April 2026) with four tabs in distinct cwds. Three consecutive jump clicks on the target session all landed on the correct tab with zero beeps, zero permission errors, and the expected "Focused the matching Warp tab." return message.

## Known limitations (follow-up work)

- **`WarpSQLiteReader.lookupPaneUUID` falls through when shell never rendered a prompt**: If a user launches an agent via a compound command that skips the shell prompt (e.g. `mkdir -p /tmp/foo && cd /tmp/foo && claude` in one line), Warp's shell integration never OSCs the new cwd to the SQLite `terminal_panes.cwd` column, and the lookup returns nil. `jumpToWarpPane` then falls back to bare activation ("Activated Warp. No precise pane mapping available."). Workaround for users: run a non-TUI command (like `pwd` or `ls`) between `cd` and `claude` so the shell renders a prompt. Proper fix (tracked as a follow-up) is to cross-reference the `commands` table or use AX to read tab state live.
- **Ad-hoc dev binary Accessibility cdhash churn**: Every `launch-dev-app.sh` rebuild invalidates the Accessibility grant because TCC tracks ad-hoc binaries by cdhash. Tracked as a follow-up to introduce a stable self-signed dev signing identity.

## Test plan

- [x] `swift build` — clean
- [x] `swift test` — 158/158
- [x] Manual: jump to a Warp-hosted Claude session across 3 consecutive clicks, target always hit, no beeps
- [ ] Post-merge: verify with a production-signed build (where Accessibility permission persists) that the feature works identically to the manual verification done here on the dev bundle


Octane0411:
## Code Review Feedback

Overall the Warp precision jump feature is well-isolated — all Warp-specific code paths are guarded by `terminalApp == "Warp"` checks, so non-Warp users are not affected. The `inferTerminalApp` refactor to prioritize `TERM_PROGRAM` is the right direction and fixes a real env-var leakage bug.

Reviewed all four related PRs (#261, #262, #263, #266) — no issues found that would impact non-Warp users. LGTM.

### 💡 Non-blocking suggestions

1. **Pre-existing issue in `CodexHooks.swift`**: The `case "vscode"` branch in `inferTerminalApp` doesn't check `CURSOR_TRACE_ID` to differentiate Cursor from VS Code (unlike `ClaudeHooks.swift` which does). This is not a regression from this PR — it already exists on `main` — but since you're refactoring `inferTerminalApp` here anyway, it would be a good opportunity to fix it.

2. **Code duplication**: `inferTerminalApp` is nearly identical across `ClaudeHooks` and `CodexHooks`. Consider extracting it into a shared utility to reduce the risk of future divergences like the Cursor issue above.

---

PR #262, opened 2026-04-11
fix(core): WarpSQLiteReader commands-history fallback for stale terminal_panes.cwd

loop2zero:
## Summary

Covers the "Known limitation" explicitly called out in #261: precision jump silently degrades to bare Warp activation whenever the user launches an agent via a single compound command like `mkdir -p /tmp/foo && cd /tmp/foo && claude`.

Warp writes a `precmd-<session>-1` block at the first shell prompt render (binding the shell session to the pane uuid) but never updates `terminal_panes.cwd` past the initial `/Users/u`, because claude's TUI takes over before a second prompt can render. The primary `WHERE cwd IN (...)` query added in #261 misses, `lookupPaneUUID` returns nil, and `jumpToWarpPane` hits its "No precise pane mapping available" early return.

This PR adds a secondary lookup strategy: cross-reference the `commands` table for a row whose text contains `cd <target>` (both firmlink forms), take the session_id, join to a `precmd-<session_id>-*` block, and return the pane_leaf_uuid. The commands table retains the full pasted command text even when the shell never rendered a second prompt, so this recovers the mapping.

## Dependency on #261

⚠️ **This PR is stacked on top of #261** — it needs the firmlink fix and the other keystroke/hook fixes from that PR as its base. The full commit list in this PR is:

```
8810065 fix(core): WarpSQLiteReader falls back to commands history when terminal_panes.cwd is stale  ← new in this PR
7809ef4 fix(app): make jumpToWarpPane robust — wait for frontmost, drop misplaced Accessibility gate  (#261)
ddabafa fix(app): click "Switch to Next Tab" via AX instead of synthesizing Cmd+Shift+]  (#261)
078e9dd fix(core): WarpSQLiteReader matches both firmlink forms of cwd  (#261)
aff8f83 fix(hooks): use TERM_PROGRAM as authoritative terminal signal  (#261)
<10 earlier scaffolding commits from the precision jump feature branch>
```

**Merge order**: please merge #261 first. Once #261 lands, GitHub will automatically narrow this PR's diff down to the single `8810065` commit, and the remaining review surface is ~170 lines (implementation + 5 regression tests + fixture).

## Regression tests (5 new)

Pinned three subtle correctness properties I got wrong in early drafts before arriving at the final query:

1. **Substring collision** — `cd /tmp/foo` must not match a command containing `cd /tmp/foo-extended`. Pinned by `lookupPaneUUIDFallbackDoesNotFalsePositiveOnSubstringCwd`. Uses SQLite `GLOB '*cd <path>[ ;&|]*'` with a shell-separator character class to bound the match.
2. **Orphan blocks from closed tabs** — Warp prunes `terminal_panes` rows when the user closes a tab but leaves the historical `commands` and `blocks` in place. My dev machine has ~1000 orphan blocks per live pane. Without a filter the fallback would return a uuid that hasn't existed for hours, and the precision jump cycle loop would spin its entire cap chasing a ghost. Pinned by `lookupPaneUUIDFallbackSkipsOrphanBlocksFromClosedTabs`, fixed by inner-joining `terminal_panes`.
3. **Firmlink input** — hook payloads arrive with `/private/tmp/foo` via `getcwd()` but the user typed `cd /tmp/foo`. Pinned by `lookupPaneUUIDFallbackAcceptsFirmlinkFlippedInput`, reusing the existing `cwdLookupCandidates` helper.

Plus 2 sanity tests: `lookupPaneUUIDFallsBackToCommandHistoryWhenTerminalPanesCwdIsStale` (the happy path this PR exists to fix) and `lookupPaneUUIDPrimaryStillWinsWhenTerminalPanesCwdIsPopulated` (regression guard against accidentally preferring the slower fallback when the primary works).

Full suite: **163/163 green** (158 from #261 baseline + 5 new here).

## Known limitations (acceptable)

- Exotic shell separators outside `[ ;&|]` (plain newline, trailing-slash path normalization) are not handled. The space-separator case covers `cd <path> && claude`, `cd <path>; claude`, `cd <path> | tee`, etc. which account for ~100% of pasted-one-liner agent launches I've seen in practice.
- Paths containing SQLite GLOB meta characters (`*`, `?`, `[`, `]`) skip the fallback silently. Vanishingly rare; the primary lookup still handles them correctly.
- If two live Warp tabs shared the same `cd <target>` in their command history (e.g. two windows both `cd /tmp/foo`'d), the most recent `commands.id` wins. Consistent with the primary lookup's same-cwd tiebreaker.

## Test plan

- [x] `swift build`
- [x] `swift test` — 163/163
- [x] 5 new regression tests pinned above
- [ ] Post-merge: manual verification on a production-signed build with the compound-command flow (`mkdir -p /tmp/foo && cd /tmp/foo && claude`) — precision jump should now land on the correct tab without the `pwd`-ritual workaround.


loop2zero:
## Update: fallback strategy replaced after real-world testing

I just pushed `27bef73` which fundamentally changes how this fallback works. The original commit `8810065` relied on `precmd-<session>-*` blocks existing for every shell session, and used a `commands → blocks → terminal_panes` 3-way join. Real E2E testing against a live Warp SQLite surfaced the assumption was wrong.

**What I saw**: a new Warp tab opened with `mkdir -p /tmp/test-c && cd /tmp/test-c && claude` pasted as a single line had:
- `terminal_panes` row with stale cwd `/tmp/test-b` (inherited from the previously-focused sibling tab, never updated)
- `commands` row with the full compound text and `session_id` 177587767426794
- **Zero `precmd-177587767426794-*` blocks** — claude grabbed the TTY before Warp's shell integration could OSC even the first prompt

The original fallback's `JOIN blocks b ON b.block_id LIKE 'precmd-' || c.session_id || '-%'` returned nothing, so the fallback returned nil, and precision jump fell back to "Activated Warp. No precise pane mapping available" — exactly the scenario this PR was supposed to fix.

**New correlation path**: `commands.pwd = terminal_panes.cwd`. The `commands.pwd` column captures the shell's cwd at the moment a command starts, which for a compound command in a freshly-opened tab equals the pane's inherited initial cwd — which Warp DOES write to `terminal_panes.cwd` at pane creation even when no precmd fires.

**Disambiguation signals** (Warp opens new tabs inheriting cwd from the focused tab, so two sibling tabs often share the same stale initial cwd):

1. `LEFT JOIN blocks b_match ON b_match.pane_leaf_uuid = tp.uuid AND b_match.block_id GLOB 'precmd-' || c.session_id || '-*'` + `ORDER BY (b_match.id IS NOT NULL) DESC` — prefer a pane already linked via precmd block to the command's own session.
2. `NOT EXISTS (SELECT 1 FROM blocks b_foreign WHERE ... AND b_foreign.block_id NOT GLOB 'precmd-' || c.session_id || '-*')` — exclude panes owned by a DIFFERENT shell session via their own precmd blocks. Without this, the lookup for test-c would return test-b's pane (sibling stale cwd, higher id, visibly wrong because blocks tie it to test-b's session).
3. `c.id DESC, tp.id DESC` — most recent command, most recent pane as final tiebreaker.

Verified on the real Warp SQLite:
```
hex(tp.uuid)                       | c.id | session_id       | command
4B762C4C1845409287370E8C111AA9AA   | 2333 | 177587767426794  | mkdir -p /tmp/test-c && cd /tmp/test-c && claude
```

And verified end-to-end in a fully integrated local dev bundle (PR #261 + #262 + #263 merged): double-clicking the test-c session in Open Island correctly focuses the test-c Warp tab after the fix.

Added a new regression test `lookupPaneUUIDFallbackFindsPaneEvenWithoutAnyPrecmdBlocks` with a new fixture scenario (`compoundCommandInSiblingTabFlow`) that models the exact sibling-tabs-with-shared-inherited-cwd collision. The old test cases (`...FallsBackToCommandHistoryWhenTerminalPanesCwdIsStale`, `...FallbackAcceptsFirmlinkFlippedInput`, `...FallbackDoesNotFalsePositiveOnSubstringCwd`, `...FallbackSkipsOrphanBlocksFromClosedTabs`) also all still pass because the NOT EXISTS filter handles the same cases the block-join did, plus more.

Octane0411:
Closing this — all commits here are included in #266, which has been reviewed and is ready to merge. No work is lost.

PR #266, opened 2026-04-11
feat(core): pid-based Warp pane lookup for sibling-tab disambiguation

loop2zero:
## Summary

Covers the edge case that neither the primary `terminal_panes.cwd` lookup nor the commands-history fallback from #262 can resolve: **two live Warp tabs with identical cwd**.

Common real-world scenario: one tab runs a Claude/Codex agent, a sibling tab is an interactive shell cd'd to the same project dir for `git status` / `pnpm build` / etc. Both panes share a cwd, both are live, neither has a distinguishing precmd block — and the existing `ORDER BY id DESC` heuristic arbitrarily picks one. Half the time it's the wrong one; precision jump lands on the sibling.

Warp exposes **zero process-linkage fields in its SQLite schema** (I scanned all 46 tables during task #6 investigation — no pid/tty/process_id/shell_pid column anywhere). The fix can't come from inside the database. Instead, this PR correlates via process-tree topology.

## Core observation

Warp spawns one shell subprocess per pane as a direct child of its `stable terminal-server` helper. Both the shell children AND the `terminal_panes` rows are created in the same order — one pair per "open new tab" action. So:

> The Nth child of terminal-server (by pid ASC) corresponds to the Nth pane (by `tab_id ASC`).

This was verified live against the user's real Warp state during E2E debugging of #262:

```
terminal-server pid: 8375
children sorted ASC: [34809, 62839, 65300]

pane at index 0 → FCE4933185444D7BB888A7F4F79A6025 (giftcard)
pane at index 1 → 40FB69CC5F6948F0801FBA63EFE2ACB4 (test-b)
pane at index 2 → 4B762C4C1845409287370E8C111AA9AA (test-c)
```

Note that the test-c pane's `terminal_panes.cwd` was still the stale `/tmp/test-b` inherited from its sibling tab, but pid-based lookup returned the correct uuid anyway — the algorithm bypasses cwd entirely.

### `Sources/OpenIslandCore/WarpProcessResolver.swift` (new)

Walks the calling process's `getppid` chain upward until it finds a parent whose command matches Warp's `terminal-server` signature. The last pid in the chain whose parent IS terminal-server is the "pane shell pid" — uniquely identifying the Warp pane that hosts the caller. Walk depth capped at 16 (hook CLI → claude → zsh → maybe-wrapper → Warp shell comfortably fits). Default providers shell out to `/bin/ps`, but the walker takes injectable providers so tests pin the logic without subprocesses.

### `Sources/OpenIslandCore/WarpSQLiteReader.swift`

Two new overloads:

1. **`lookupPaneUUIDByShellPID(_:siblings:)`** (pure): takes a shellPID plus the siblings list, sorts ASC, finds the shellPID's index, returns the Nth pane uuid from the `tabs`-joined `terminal_panes` query. Unit-testable without touching the process table.
2. **`lookupPaneUUIDByShellPID(_:terminalServerPID:)`** (production wrapper): enumerates children via `/usr/bin/pgrep -P` and delegates to the pure variant.

Query structure drops the `active_window_id` filter that `currentFocusedPaneUUID` uses, because on real hardware that column is sometimes NULL (Warp populates it lazily) — filtering on it returns zero rows even though the panes are plainly live. The pid-correlation assumption holds across all windows, so ordering by `tab_id ASC` without window filtering is both correct and simpler.

### `Sources/OpenIslandCore/ClaudeHooks.swift` / `CodexHooks.swift`

The default `warpPaneResolver` closure now tries the pid-based lookup first, falls back to the existing cwd-based path from #262 when the hook is not inside a Warp pane (test invocations, Ghostty-hosted agents, Warp terminal-server restart etc.). No public interface changes — the resolver is still a `(String) -> String?` closure, only the default implementation changed.

## Tests

- **`WarpProcessResolverTests`** (8 tests): pin the tree walker's happy path, Ghostty-hosted negative case, depth cap against cyclic chains, case-insensitive match for the terminal-server signature, rejection of Warp's main GUI process, and the "must not report Warp main as terminal-server" invariant.
- **`WarpSQLiteReaderTests`** (4 new): pin the index arithmetic for in-range lookups, out-of-range rejection, absent-pid rejection, and caller-supplied-out-of-order siblings.

Full suite: **176/176** (previous baseline 163 + 13 new here).

## Dependency

⚠️ **Stacks on top of #262.** GitHub will narrow this PR's diff automatically once #262 merges. The single new commit in this PR is `90faaa0`.

## Known limitations

- **Warp terminal-server restart**: child pids reset but terminal_panes.id persists, breaking the position correlation until the next rebuild. In practice terminal-server only restarts during a Warp upgrade. The cwd-based fallback from #262 still catches the common case.
- **Warp pane splits**: untested. If Warp supports split panes and the split creates extra `terminal_panes` rows that are NOT direct children of terminal-server, the index arithmetic would drift. Falls back to cwd-based lookup gracefully.
- **Subprocess cost**: pid-based path adds 2 `ps` calls (chain walk) + 1 `pgrep -P` (enumerate children) per hook event. Measured under 10 ms total on M-series silicon — negligible.

## Test plan

- [x] `swift build` — clean
- [x] `swift test` — 176/176
- [x] Unit tests pin the index correlation arithmetic
- [x] Live verification against real Warp SQLite (3 tabs, 3 shell pids, 3 panes — all resolve correctly in both algorithms)
- [ ] Post-merge E2E: open two Warp tabs at the same cwd, confirm precision jump lands on the correct one. This case is hard to reproduce automatically but is the exact failure mode this PR exists to fix.


loop2zero:
## Update: Codex review fixes pushed (`c576042`)

All 3 findings from the Codex review above are addressed. Summary:

### [P1] Multi-window background-pane rejection

`lookupPaneUUIDByShellPID` now pulls `t.window_id` alongside the pane uuid and rejects the row in Swift when the pane lives in a non-active window — the cycle loop in `jumpToWarpPane` only cycles the frontmost window, so returning a background-window uuid would spin its entire cap before timing out. When `app.active_window_id` reads back NULL (observed live on real hardware), a new `fetchActiveWindowID(db:)` helper falls back to `MIN(window_id) FROM tabs` so single-window users don't regress.

New regression tests:
- `lookupPaneUUIDByShellPIDReturnsNilWhenPaneIsInBackgroundWindow`
- `lookupPaneUUIDByShellPIDFallsBackToSmallestWindowWhenActiveWindowIsNull`

### [P1] Split-tab offset drift in `currentFocusedPaneUUID`

The legacy query offset into `tabs JOIN pane_nodes` using `active_tab_index`, which counts tabs but the join returns one row per LEAF — so any earlier split inflated the row set and shifted the offset. Decomposed the query into three explicit Swift-driven SQL steps:

1. Resolve `activeWindowID` (with NULL fallback, shared helper)
2. Resolve `activeTabIndex` for that window
3. Resolve the `tab_id` at that offset within `tabs` only (no join)
4. Look up the leaf pane of that specific tab, preferring `pane_leaves.is_focused=1` when set

**Sanity-check discovery during verification**: running the legacy query against the current live Warp SQLite actually fails with `stepping, datatype mismatch (20)` — `app.active_window_id` is NULL right now and the nested `OFFSET (SELECT … WHERE id = NULL)` propagates NULL into OFFSET. So the old implementation has been silently broken for an unknown number of Warp states; this PR's fix is the first time `currentFocusedPaneUUID` actually returns a uuid for the current Warp state.

One nuance worth flagging: on real Warp databases `pane_leaves.is_focused = 1` for every row (Warp appears to set it unconditionally on leaf creation and never clear it). The `ORDER BY is_focused DESC` preference effectively decays into the `pn.id ASC` tiebreaker for split tabs. Single-leaf tabs always return correctly; split tabs return a deterministic-but-possibly-not-focused leaf the user can adjust manually. Documented inline.

New regression tests:
- `currentFocusedPaneUUIDReturnsFocusedLeafOfActiveTabWithSplits`
- `currentFocusedPaneUUIDPrefersTheFocusedLeafWithinASplitTab`
- `currentFocusedPaneUUIDRecoversWhenActiveWindowIDIsNull`

### [P2] Transient resolver nil no longer wipes `warpPaneUUID`

Extracted the jumpTarget merge logic out of both `synchronizeClaudeJumpTarget` and `synchronizeJumpTarget` (Codex) into a shared static helper `mergeJumpTargetPreservingExistingResolvedFields(incoming:existing:)`. The helper preserves both `terminalSessionID` (the pre-existing behavior, now deduplicated) and `warpPaneUUID` (new) when the incoming field is nil and the existing one has a value. Called from both sync functions.

The helper is `internal` (default Swift visibility) so the behavior can be pinned with pure unit tests, no BridgeServer socket dance required.

New regression suite: `BridgeServerJumpTargetMergeTests` with 5 cases covering nil preservation, successful overwrite, both-sides-nil, and missing-existing-jumpTarget.

### Test counts

- Before this commit: 176 tests in 20 suites.
- After this commit: **186 tests in 21 suites** (+10, all green).

### Live verification

Walked the new `currentFocusedPaneUUID` steps against the live database manually:

```
active_window_id: [NULL]
MIN(window_id) fallback: 1
active_tab_index for window 1: 1
tab at offset 1 in window 1: 2
leaf pane of tab 2: 40FB69CC5F6948F0801FBA63EFE2ACB4 (cwd /tmp/test-b)
```

The old query errored out at this exact state. The new one returns the correct uuid.

Octane0411:
CI `harness` job is failing:

```
docs index is missing link to: docs/exec-plans/active/2026-04-10-warp-precision-jump-design.md
```

The new design doc needs to be registered in the docs index, or removed from the PR if it's only meant as internal working notes. We have strict docs hygiene enforced by CI — every doc under `docs/` must be linked in the index.

Please fix and push, or let me know if you'd prefer to drop the exec-plan docs from this PR entirely.

---

PR #328, opened 2026-04-15
feat+fix: wrapper-mode usage bridge, visible hook sessions, reliable sound

DoTheWorkNow:
## Summary / 概述

This PR bundles one new feature and four fixes that surfaced when running the
app from a fresh fork install. Each commit stands alone and targets a
separate symptom. All 203 unit tests pass.

本 PR 合并了一项新功能和四处修复，都是作者从新 fork 本地跑起来时遇到的问题，每个 commit 都可单独 review。全部 203 个单元测试通过。

---

### 1. `feat: install Claude usage bridge in wrapper mode when statusLine is occupied` (ee49078)

**Problem.** The "Install" button under Settings → Claude usage bridge throws
`existingStatusLineConflict` and refuses to do anything if the user already has
a custom `statusLine.command` in `~/.claude/settings.json` (e.g. claude-hud).
Many Claude Code users have one — they get a dead button with no useful
feedback.

**Fix.** Added `installAsWrapper()`, which:

1. Saves the user's original `statusLine` dict verbatim under a new
   `_openIslandOriginalStatusLine` key in settings.json.
2. Writes `~/.open-island/bin/open-island-statusline` (the wrapper) and
   `open-island-statusline-delegate` (contains the user's original command).
3. Points `statusLine.command` at the wrapper.

The wrapper reads Claude Code's stdin once, writes `.rate_limits` to the
cache file, and pipes the same stdin through the delegate script so the
user's custom statusline keeps rendering. `uninstall()` restores the
original from `_openIslandOriginalStatusLine` verbatim.

`HookInstallationCoordinator.installClaudeUsageBridge()` catches
`existingStatusLineConflict` and transparently falls back to wrapper
mode, so the existing UI button "just works" without changing the UX.

Two new tests cover the wrapper install/uninstall round-trip and the
auto-fallback behavior.

**中文。** 用户原先有自定义 statusLine（如 claude-hud）时，装用量桥接会静默失败。新增 wrapper 模式：保留原脚本、在外面套一层 shim 同时写 rate_limits 缓存。装/卸都能完整还原。

---

### 2. `fix: generalize empty-state prompt to not hardcode Codex` (d0614d4)

**Problem.** When no sessions are visible, the island shows
\"Start Codex in your terminal\" — regardless of whether the user even uses
Codex. With a Claude-only setup, seeing a Claude usage header above a
\"Start Codex\" hint is misleading.

**Fix.** All three locales now say \"Start a coding agent in your terminal\"
(\"在终端中启动编码 Agent\" / \"在終端機中啟動編碼 Agent\").

**中文。** 空状态硬编码 \"Codex\" 对只用 Claude Code 的用户很奇怪，改成通用文案。

---

### 3. `fix: keep hook-managed sessions visible when process discovery can't match them` (ab154b1)

**Problem.** Active hook-managed Claude Code sessions would vanish from the
island list even though hook events were flowing. Permission prompts would
flash on screen and then disappear on their own a few seconds later, before
the user could click Yes / No — visually this made it look like hovering the
card caused it to close.

**Root cause.** `ActiveAgentProcessDiscovery` tries to pair each state session
to a live OS process. Claude Code's Node process doesn't expose the session
id on its command line, so the heuristic misses frequently. After **two**
consecutive misses, `SessionState.updateLiveness(aliveSessionIDs:)` was
stamping the session with `isSessionEnded=true, phase=.completed`, which
drops it out of `isVisibleInIsland`. Nothing in the activity / permission /
question reducers was clearing those flags when a later hook event proved the
agent was still running. For permission prompts the damage was worse: the
card would show briefly because `phase.requiresAttention` bypasses the
visibility check, then the very next tick flipped phase back to `.completed`
and nuked the card.

**Fix.** Two changes in `SessionState.swift`:

- `.activityUpdated`, `.permissionRequested`, `.questionAsked` reducers now
  clear `isSessionEnded` and reset `processNotSeenCount = 0`. An incoming
  hook event is authoritative proof the agent is alive.
- The hook-managed branch of `updateLiveness(aliveSessionIDs:)` refuses to
  mark a session ended while `phase.requiresAttention` — a stale discovery
  miss must never remove a prompt the user still has to answer.

**中文。** hook 管理的会话被进程发现误杀，导致权限卡片一闪就消失。修复 reducer 上所有活动事件复活 session，且对正在等待用户响应的会话禁止被清理。

---

### 4. `fix: raise hook-managed process-miss threshold and reset it on Stop hook` (f174ec3)

**Problem.** Follow-up to #3. With only that fix, an idle Claude Code
session between turns (phase=`.completed` after Stop hook) would still be
killed within seconds: Stop is "turn complete," not "session ended," yet the
next process-discovery tick saw phase=.completed and went ahead with
`isSessionEnded=true`.

**Fix.**

- `.sessionCompleted` reducer resets `processNotSeenCount = 0` unless the
  payload has `isSessionEnd=true`. Claude's Stop hook carries
  `isSessionEnd=nil`, so idle-between-turns now stays alive.
- Raise the hook-managed miss threshold from **2** to **30**. It's still a
  last-resort crash-detection fallback, but discovery hiccups no longer
  cost a live session.

**中文。** Stop hook 是"回合结束"不是"会话结束"，原先每次 Stop 后进程发现仍可能把 session 打死。阈值从 2 放宽到 30，且 Stop 重置计数。

---

### 5 + 6. `fix: sound — NSSound → AudioServices → afplay` (8cbe344, 89941dd)

**Problem.** Notification sounds don't play when the app is launched from
`swift run OpenIslandApp` — even with mute off, volume up, and the sound
file present. `NSSound(named:).play()` returns a valid NSSound instance
and `play()` returns true, but the speakers stay silent.

**Investigation.** Started with the obvious suspect: NSSound's bundle
dependency. Rewrote to `AudioServicesPlaySystemSound` (lower-level C API,
supposedly no bundle coupling). On macOS 26.3 that **also** silently
no-ops from an unbundled context: `AudioServicesCreateSystemSoundID`
returns `kAudioServicesNoError`, `AudioServicesPlaySystemSound` runs, and
nothing plays. Both APIs route through HIServices / CoreAudio's
per-process registration which quietly refuses un-registered hosts.

**Fix.** Shell out to `/usr/bin/afplay` on a background queue. It's a
tiny CLI shipped with every macOS install, has no bundle dependency,
respects system volume, and was ground-truth evidence earlier that the
audio stack itself was healthy. Fire-and-forget — we don't wait.

Left both commits in the history so the investigation trail is
reviewable. A squash-merge collapses them to one line if preferred.

**中文。** `swift run` 起的 App 里 NSSound 和 AudioServices 都会静默失败，原因是 macOS 的音频注册机制要求正规 .app bundle。改为 fork `/usr/bin/afplay` 子进程，绕开该限制。

---

## Test plan / 测试

- [x] `swift build` — clean
- [x] `swift test` — all 203 tests pass (includes 2 new tests for wrapper mode)
- [x] Manual: install wrapper mode while claude-hud is the existing
      statusLine; confirmed both the claude-hud line and the Open Island
      usage header render correctly in a fresh Claude Code turn
- [x] Manual: Claude sessions across two different cwds (with and without
      `--dangerously-skip-permissions`) stay visible during idle between
      turns and don't vanish on hover
- [x] Manual: notification sound plays reliably on Stop hook from
      `swift run` build on macOS 26.3

## Summary by CodeRabbit

* **New Features**
  * Added wrapper installation mode for existing status line configurations, preserving them instead of failing installation.

* **Bug Fixes**
  * Improved session stability and crash detection for managed processes.

* **Documentation**
  * Updated UI terminology from "Codex" to "coding agent" across English, Simplified Chinese, and Traditional Chinese localizations.

* **Tests**
  * Added test coverage for wrapper installation scenarios.

### DIFF
--- a/Sources/OpenIslandApp/HookInstallationCoordinator.swift
+++ b/Sources/OpenIslandApp/HookInstallationCoordinator.swift
@@ -888,7 +888,13 @@ final class HookInstallationCoordinator {
 
     func installClaudeUsageBridge() {
         updateClaudeUsageBridge(userMessage: "Installing Claude usage bridge.") { manager in
-            try manager.install()
+            do {
+                return try manager.install()
+            } catch ClaudeStatusLineInstallationError.existingStatusLineConflict {
+                // User already has a custom statusLine (e.g. claude-hud). Install as a
+                // wrapper so their script keeps running and we still get rate_limits.
+                return try manager.installAsWrapper()
+            }
         }
     }
 
@@ -1073,7 +1079,11 @@ final class HookInstallationCoordinator {
                 self.claudeStatusLineStatus = status
                 self.claudeUsageSnapshot = try ClaudeUsageLoader.load()
                 if status.managedStatusLineInstalled {
-                    self.onStatusMessage?("Claude usage bridge is installed. Start a Claude Code turn to refresh cached rate limits.")
+                    if status.managedStatusLineIsWrapper {
+                        self.onStatusMessage?("Claude usage bridge installed in wrapper mode — your existing statusLine is preserved. Start a Claude Code turn to refresh cached rate limits.")
+                    } else {
+                        self.onStatusMessage?("Claude usage bridge is installed. Start a Claude Code turn to refresh cached rate limits.")
+                    }
                 } else {
                     self.onStatusMessage?("Claude usage bridge is not installed.")
                 }
--- a/Sources/OpenIslandApp/NotificationSoundService.swift
+++ b/Sources/OpenIslandApp/NotificationSoundService.swift
@@ -1,11 +1,20 @@
 import AppKit
+import AudioToolbox
 
 /// Manages notification sound playback using macOS system sounds.
+///
+/// Uses AudioToolbox's `AudioServicesPlaySystemSound` instead of `NSSound`
+/// because the latter relies on bundle context and silently fails when the
+/// app is launched via `swift run` or any other non-bundled path.
 @MainActor
 struct NotificationSoundService {
     private static let soundsDirectory = "/System/Library/Sounds"
     private static let defaultsKey = "notification.sound.name"
     static let defaultSoundName = "Bottle"
+
+    /// Cache of registered system sound IDs so we don't re-register the same
+    /// file on every notification. Keyed by sound name (e.g. "Bottle").
+    private static var soundIDCache: [String: SystemSoundID] = [:]
 
     /// Returns the list of available system sound names (without file extension).
     static func availableSounds() -> [String] {
@@ -31,11 +40,24 @@ struct NotificationSoundService {
 
     /// Plays a system sound by name.
     static func play(_ name: String) {
-        guard let sound = NSSound(named: NSSound.Name(name)) else {
-            return
+        let soundID: SystemSoundID
+        if let cached = soundIDCache[name] {
+            soundID = cached
+        } else {
+            let url = URL(fileURLWithPath: soundsDirectory)
+                .appendingPathComponent("\(name).aiff")
+            guard FileManager.default.fileExists(atPath: url.path) else {
+                return
+            }
+            var id: SystemSoundID = 0
+            let status = AudioServicesCreateSystemSoundID(url as CFURL, &id)
+            guard status == kAudioServicesNoError else {
+                return
+            }
+            soundIDCache[name] = id
+            soundID = id
         }
-        sound.stop()
-        sound.play()
+        AudioServicesPlaySystemSound(soundID)
     }
 
     /// Plays the user-selected notification sound, respecting the mute setting.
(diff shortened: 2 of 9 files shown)


Octane0411:
Thanks for your work, Could you split this into separate PRs per topic since the commits are already independent.

DoTheWorkNow:
Thanks @Octane0411 for the review — split as requested into one PR per topic. The commits are already independent so no code changes besides rebasing onto upstream/main and addressing @coderabbitai's nitpick about mixing `Agent` (English) into the Chinese locales (now fully localized as `编码代理` / `編碼代理`).

- #330 feat: install Claude usage bridge in wrapper mode when statusLine is occupied
- #331 fix: generalize empty-state prompt to not hardcode Codex (also folds the coderabbit nitpick)
- #332 fix: keep hook-managed sessions visible when process discovery can't match them (2 commits)
- #333 fix: play notification sounds reliably from unbundled dev builds (2 commits)
- #334 chore: disable Sparkle auto-update check in DEBUG builds

Closing this one — happy to re-land any piece if a specific split would land better.

PR #333, opened 2026-04-15
fix: play notification sounds reliably from unbundled dev builds

DoTheWorkNow:
Split out of #328.

## Problem

Notification sounds don't play when the app is launched from `swift run OpenIslandApp` — even with mute off, volume up, and the sound file present. `NSSound(named:).play()` returns a valid NSSound instance and `play()` returns true, but the speakers stay silent.

## Investigation

Started with the obvious suspect: NSSound's bundle dependency. Rewrote to `AudioServicesPlaySystemSound` (lower-level C API, supposedly no bundle coupling). On macOS 26.3 that **also** silently no-ops from an unbundled context — `AudioServicesCreateSystemSoundID` returns `kAudioServicesNoError`, `AudioServicesPlaySystemSound` runs, and nothing plays. Both APIs route through HIServices / CoreAudio's per-process registration which quietly refuses un-registered hosts.

## Fix

Two commits, kept separate so the investigation trail is reviewable (squash-merge collapses them fine):

1. **AudioServices switch** — addresses the NSSound bundle-lookup issue in principle.
2. **afplay fallback** — shells out to `/usr/bin/afplay` on a background queue when AudioServices doesn't actually emit audio. `afplay` ships with every macOS install, has no bundle dependency, respects system volume, and was ground-truth evidence earlier that the audio stack itself was healthy. Fire-and-forget.

`swift run` 下 NSSound 和 AudioServices 都会被 macOS 音频注册机制拒绝；改为 fork `/usr/bin/afplay` 子进程，绕开所有 bundle 限制。

## Test plan

- [x] `swift build` / `swift test` (203 passing)
- [x] Manual: notification sound plays reliably on Stop hook from `swift run` build on macOS 26.3.

### DIFF
--- a/Sources/OpenIslandApp/NotificationSoundService.swift
+++ b/Sources/OpenIslandApp/NotificationSoundService.swift
@@ -1,11 +1,20 @@
 import AppKit
+import AudioToolbox
 
 /// Manages notification sound playback using macOS system sounds.
+///
+/// Uses AudioToolbox's `AudioServicesPlaySystemSound` instead of `NSSound`
+/// because the latter relies on bundle context and silently fails when the
+/// app is launched via `swift run` or any other non-bundled path.
 @MainActor
 struct NotificationSoundService {
     private static let soundsDirectory = "/System/Library/Sounds"
     private static let defaultsKey = "notification.sound.name"
     static let defaultSoundName = "Bottle"
+
+    /// Cache of registered system sound IDs so we don't re-register the same
+    /// file on every notification. Keyed by sound name (e.g. "Bottle").
+    private static var soundIDCache: [String: SystemSoundID] = [:]
 
     /// Returns the list of available system sound names (without file extension).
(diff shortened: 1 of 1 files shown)


Octane0411:
Thanks for the investigation! One question — our recommended dev workflow uses scripts/launch-dev-app.sh (bundled context), where NSSound works fine. Is there a scenario where swift run is preferred and this fix becomes necessary?

DoTheWorkNow:
You're right — just verified with a bundle rebuilt via `scripts/launch-dev-app.sh` (NSSound version, unchanged from main), and the Stop-hook notification plays fine. The failure I hit was specific to `swift run OpenIslandApp` for quick iteration. Since that's not your recommended dev path, this PR isn't needed. Closing.

Thanks for the thoughtful question — I'd missed that the dev-signed bundle from `launch-dev-app.sh` is effectively a "real" bundle and NSSound registers correctly there.

PR #369, opened 2026-04-19
fix(opencode): overhaul process discovery and session liveness tracking

Raygooo:
- Refine `isOpenCodeProcess` to parse argv tokens instead of relying on broad substring matches, preventing unrelated processes (e.g., `vim ~/.opencode/...`) from keeping sessions alive.
- Add robust support for package runners (`npx`, `npm exec`, `yarn`, `pnpm dlx`) executing the `opencode` CLI.
- Deduplicate process snapshots by TTY and working directory rather than PID. This ensures that a package runner and its child Node process don't produce duplicate snapshots.
- Overhaul OpenCode session liveness in `ProcessMonitoringCoordinator` by uniquely matching active processes to tracked sessions via their TTY and normalized working directories.
- Eliminate the previous fallback that blindly marked all OpenCode sessions alive if any OpenCode process existed.
- Prevent claiming a session when the TTY matches but the working directory explicitly differs (e.g., terminal tab was reused for a different workspace).
- Fallback to keeping remaining unclaimed sessions alive only if there are active OpenCode processes that couldn't be uniquely matched, preventing accidental session termination.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate OpenCode process detection, including launches via package runners and explicit CLI/node invocations.
  * Consistent capture of process working directories in snapshots for improved context.
  * Reduced duplicate snapshots by deduplicating based on terminal and working directory information.
  * Populate terminal application info and resolve tmux-derived details on snapshots when available.
  * Refined per-process session liveness matching to avoid incorrectly ending or duplicating OpenCode sessions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ build/
 
 # Swift Package Manager
 .build/
+.swiftpm/
 Packages/
 Package.resolved
 
@@ -24,3 +25,4 @@ output/
 *.log
 *.tmp
 .worktrees/
+.air/
--- a/Sources/OpenIslandApp/ActiveAgentProcessDiscovery.swift
+++ b/Sources/OpenIslandApp/ActiveAgentProcessDiscovery.swift
@@ -100,17 +100,55 @@ struct ActiveAgentProcessDiscovery {
             }
 
             if isOpenCodeProcess(command: process.command) {
-                let claimKey = "opencode:\(process.pid)"
+                let lsofOutput = lsofOutput(pid: process.pid)
+                let cwd = lsofOutput.flatMap(workingDirectory(from:))
+
+                // Deduplicate by TTY and working directory instead of PID.
+                // Wrappers like `npm exec` and their child `node` process share the same TTY and CWD.
+                // Treat a missing cwd as a wildcard for the TTY to match existing claims, but let concrete cwd overwrite wildcard.
+                let ttyId = process.terminalTTY ?? process.pid
+                
+                if cwd == nil {
+                    // Only insert wildcard if no concrete claim exists for this TTY.
+                    if claimedKeys.contains(where: { $0.hasPrefix("opencode:\(ttyId):") && $0 != "opencode:\(ttyId):" }) {
+                        continue
+                    }
+                } else {
+                    // Concrete CWD: remove any existing wildcard claim for this TTY before claiming.
+                    if claimedKeys.remove("opencode:\(ttyId):") != nil {
+                        // Also remove the orphaned snapshot that was previously appended.
+                        snapshots.removeAll {
+                            $0.tool == .openCode && $0.terminalTTY == process.terminalTTY && $0.workingDirectory == nil
+                        }
+                    }
+                }
+
+                let claimKey = "opencode:\(ttyId):\(cwd ?? "")"
                 guard claimedKeys.insert(claimKey).inserted else {
                     continue
                 }
 
-                snapshots.append(ProcessSnapshot(
+                var snapshot = ProcessSnapshot(
                     tool: .openCode,
                     sessionID: nil,
-                    workingDirectory: nil,
-                    terminalTTY: process.terminalTTY
-                ))
+                    workingDirectory: cwd,
+                    terminalTTY: process.terminalTTY,
+                    terminalApp: terminalApp(for: process, processesByPID: processesByPID)
+                )
+
+                if snapshot.terminalApp == nil, let agentTTY = process.terminalTTY {
+                    if let (tmuxTarget, hostTerminalApp, socketPath) = resolveTmuxInfo(
+                        agentTTY: agentTTY,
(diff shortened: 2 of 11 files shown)


Raygooo:
### PR 总结 (中文版)

这次提交重构了 OpenCode 的进程发现与会话存活状态监控逻辑：

- **改进 `isOpenCodeProcess` 逻辑**：不再使用宽泛的子串匹配，而是通过解析 argv 的 Token 来判断，防止不相关的进程（例如 `vim ~/.opencode/...`）错误地将 OpenCode 会话保持在活跃状态。
- **支持包管理器运行**：增加对包管理器（`npx`、`npm exec`、`yarn`、`pnpm dlx`）执行 `opencode` CLI 的稳健支持。
- **改进进程去重机制**：依据终端 TTY 和工作目录（而非 PID）对进程快照进行去重。这确保了包管理器及其子 Node 进程不会产生导致异常的重复快照。
- **重构存活状态监控**：优化了 `ProcessMonitoringCoordinator` 中的 OpenCode 会话存活状态监控逻辑，通过 TTY 和标准化后的工作目录，将活跃进程唯一匹配到受监控的会话上。
- **移除粗暴的保活逻辑**：去除了原先“只要存在任何 OpenCode 进程，就盲目将所有 OpenCode 会话标记为活跃”的错误兜底逻辑。
- **防范终端标签页复用**：修复了 TTY 匹配但工作目录明确不同时，错误认领旧会话的问题（例如，在不同的工作区复用了同一个终端标签页）。
- **安全的兜底策略**：引入了新的安全兜底机制——只有在存在无法被唯一匹配的活跃 OpenCode 进程时，才将剩余未认领的会话保留为存活状态，防止真实会话被意外终止。

Raygooo:
### Address CodeRabbit Review

Thanks for the precise review! I have pushed a follow-up commit to address both edge cases:

1. **Robust Package Runner Parsing**: Updated `isOpenCodeProcess` to safely handle absolute paths (e.g. `/npm`, `/pnpm`) and strip off version suffixes like `@latest` before matching the package name, while still correctly preserving scoped packages like `@opencode-ai/...`.
2. **Tri-State Conflict Resolution**: Replaced the binary return value of `uniqueTrackedOpenCodeSession` with an `OpenCodeMatchResult` enum (`.matched`, `.ambiguous`, `.rejectedConflict`). Now, if the TTY matches but the CWD explicitly conflicts (e.g. terminal tab reused in another directory), it returns `.rejectedConflict` and the process is dropped without triggering the broad fallback that would accidentally keep unrelated sessions alive.

Raygooo:
### Second Round Fixes Addressed

Thanks again for the rigorous review! The latest commit addresses the final two edge cases highlighted by the AI reviewers:

1. **Precision in Install Command Checks**: The check for package-management commands (e.g. `install`, `add`) is now position-aware. We find the index of the `opencode` package token and only treat the command as an install operation if the install-like token appears *before* it. This ensures that commands like `npm run my-script opencode install` are not wrongly excluded, whereas `npm install opencode` correctly is.
2. **Wildcard TTY Deduplication via nil CWD**: The process snapshot deduplication logic now treats a missing CWD as a wildcard for that TTY. If a process emits a snapshot without a CWD but shares a TTY with a snapshot that *does* have a CWD (or vice versa), they correctly resolve as claims for the same underlying session environment, instead of spawning orphaned duplicates.

Everything is compiled and looking green! 🚀

Raygooo:
### Third Round Fixes Addressed

Thanks again for the precise catch! I've addressed the scoped paths in the child-process fallback and fixed the wildcard CWD deduplication as recommended:

1. **Scoped Package Paths in Node Sub-processes**: Updated the `isNode` fallback to correctly detect scoped `@opencode-ai` binaries and installation directories. A command invoking `.../node_modules/@opencode-ai/cli/...` or `.../.bin/@opencode-ai` will now be correctly classified as an active OpenCode process, mirroring the existing unscoped checks.
2. **Prioritizing Concrete CWDs in Process Deduplication**: Rewrote the wildcard CWD claim logic so that a snapshot with a concrete CWD takes precedence over an earlier snapshot lacking one (nil CWD). Now, if a nil-CWD wrapper is scanned first, it inserts a wildcard claim; when its child node process (with a concrete CWD) is scanned later in the same TTY, it actively removes the wildcard claim and claims the concrete CWD instead. This guarantees downstream matchers get the highest fidelity data.

Raygooo:
### Fix Applied: Orphaned Wildcard Snapshot

Addressed the critical logic flaw pointed out in the review: When a concrete-CWD process supersedes an earlier wildcard (nil-CWD) process on the same TTY, we now explicitly remove the orphaned wildcard snapshot from the `snapshots` array. Previously, leaving it in the array resulted in two snapshots being emitted for the same session environment, which defeated the deduplication and caused the `hasUnmatchedOpenCodeProcess` fallback to incorrectly trigger.

Octane0411:
thx for your contribution

PR #375, opened 2026-04-19
fix(claude): keep hook-managed sessions alive and suppress phantom synthetics when lsof transiently fails

Octane0411:
## Summary

Fixes a bug where the island's Claude session list would suddenly "collapse" — all populated sessions losing their task metadata, appearing as duplicated workspace-name-only rows, and every row stamped \"<1m\".

### Root cause

Every 2s `ProcessMonitoringCoordinator.reconcileSessionAttachments` tries to match running Claude processes to tracked hook-managed sessions via three passes:

1. `process.sessionID == session.id` (sessionID sourced from lsof / command line)
2. `process.transcriptPath == session.claudeMetadata.transcriptPath` (also lsof-sourced)
3. Unique TTY + CWD match

Both (1) and (2) depend on `lsof` returning within its 0.2s budget (`ActiveAgentProcessDiscovery.lsofCommandTimeout`). Under load lsof can return late or with partial output, dropping `sessionID` and `transcriptPath`. Pass 3 then has to carry the load, but its uniqueness check gives up as soon as multiple Claude sessions share a workspace (a common multi-terminal-on-same-repo setup).

The cascade:

1. `sessionIDsWithAliveProcesses` omits the real sessions.
2. `SessionState.markProcessLiveness` increments `processNotSeenCount`; after 2 polls it flips `isSessionEnded = true` and `removeInvisibleSessions` evicts the session.
3. `representedClaudeProcessKeys` leaves the process unrepresented, so `mergedWithSyntheticClaudeSessions` spawns a synthetic row with `updatedAt = .now`. Each subsequent tick refreshes `updatedAt` to now again — the \"<1m\" symptom.

### Fix

Add a workspace-scoped fallback pass in both `sessionIDsWithAliveProcesses` (Pass 4) and `representedClaudeProcessKeys`. When a tracked hook-managed Claude session shares a CWD with any running Claude process:

- Keep the session alive (don't tick toward `.isSessionEnded`).
- Treat the process as represented (don't spawn a phantom synthetic).

Hooks remain authoritative for phase changes; this fallback only prevents eviction / phantom rows during the matching blind-spots that happen when lsof stalls.

### Before / after

Before: user's island showed 7+ rows like:
- \`osg-scp-mds\` · Claude Code · Ghostty · <1m  (×4)
- \`osg-scp-projects\` · Claude Code · Ghostty · <1m  (×3)

All metadata (initial prompt, task, activity) gone; \"<1m\" refreshing every 2s.

After: matching transients don't evict sessions; the real hook-managed sessions stay on screen with their metadata and correct age badges.

## Test plan

- [x] `swift build` — passes
- [x] `swift test` — all 223 tests in 24 suites pass
- [x] New regression test `mergedWithSyntheticClaudeSessionsDoesNotDuplicateWhenLsofCannotDisambiguateWorkspace` — simulates lsof failing to expose sessionID/transcriptPath for multiple Claude processes in the same workspace; asserts no synthetics are spawned.
- [x] New regression test `sessionIDsWithAliveProcessesKeepsHookManagedClaudeSessionsAliveByWorkspace` — asserts hook-managed sessions stay alive when a process with matching CWD is running, even with no TTY / sessionID from lsof.
- [ ] Manual: run in dev app, open 3+ Claude sessions in the same repo, leave running for >10 min, confirm no collapse.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session liveness tracking by adding workspace-based fallback matching. Sessions now remain active in more scenarios where standard identification methods are unavailable.

* **Tests**
  * Added regression tests for workspace-based session fallback behavior.

### DIFF
--- a/Sources/OpenIslandApp/ProcessMonitoringCoordinator.swift
+++ b/Sources/OpenIslandApp/ProcessMonitoringCoordinator.swift
@@ -316,6 +316,28 @@ final class ProcessMonitoringCoordinator {
             claimedSessionIDs.insert(matched.id)
         }
 
+        // Pass 4: workspace (CWD) fallback. lsof can transiently fail to
+        // expose session ID / transcript path under load, and Pass 3's
+        // uniqueness check drops matches when multiple sessions share a
+        // workspace. Without a fallback, the hook-managed liveness timer
+        // in SessionState.markProcessLiveness would mark perfectly healthy
+        // sessions as ended after 2 polls and synthetic placeholders would
+        // take over. Keep any unclaimed tracked Claude session alive as
+        // long as at least one Claude process is running in its workspace;
+        // hooks remain authoritative for phase changes.
+        let claudeProcessCWDs: Set<String> = Set(
+            claudeProcesses.compactMap { normalizedPathForMatching($0.workingDirectory) }
+        )
+        if !claudeProcessCWDs.isEmpty {
+            for session in trackedClaudeSessions where !claimedSessionIDs.contains(session.id) {
+                guard let sessionCWD = normalizedPathForMatching(session.jumpTarget?.workingDirectory),
+                      claudeProcessCWDs.contains(sessionCWD) else {
+                    continue
+                }
+                aliveIDs.insert(session.id)
+            }
+        }
+
         // OpenCode sessions are hook-managed, but OpenCode does not expose a stable
         // session ID through process discovery. Match each active OpenCode process
         // to at most one tracked session.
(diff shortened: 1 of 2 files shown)


Octane0411:
Closing in favour of #377, which addresses the root cause instead of patching the symptom.

The CWD-scoped fallback added in this PR was a targeted workaround for the matching-path evicting healthy sessions under lsof stress. #377 removes the `ps`/`lsof` eviction path entirely for hook-managed Claude sessions and replaces it with a kernel-level `DispatchSource` process-exit monitor anchored to the Claude CLI's PID (plumbed through the hook payload as `agent_pid`). With that in place the fallback is no longer needed — on the refactor branch there's nothing to remove because the eviction path itself is gone.

The original bug scenario (`ProcessMonitoringCoordinator` returning `[]` for N consecutive reconcile cycles) is reproduced and asserted fixed in #377's `ClaudeSessionLifecycleIntegrationTests.hookManagedClaudeSessionsSurviveEmptyProcessDiscoveryCycles`.

Thanks for the patch — it kept the branch usable while the refactor was planned and shipped.

PR #398, opened 2026-04-22
fix: wire Launch at Login toggle to SMAppService (#392)

Octane0411:
## Summary
- The Launch at Login toggle in Settings -> General was bound to a view-local `@State`, so the toggle reset every time the user switched tabs and no login item was ever registered (#392).
- Back the toggle with an `AppModel.launchAtLoginEnabled` property driven by `SMAppService.mainApp`. Status is read from the system on init so the UI stays in sync even when the user toggles login items from System Settings.
- Wrap the register/unregister calls in a small `LaunchAtLoginService`. If the OS rejects the request (e.g. bundle not found when running via `swift run`), surface the error via `lastActionMessage` and revert the toggle.

## Test plan
- [x] `swift build` — clean
- [x] `swift test` — 222 tests / 24 suites pass
- [ ] Manual: launch `Open Island Dev.app`, toggle Launch at Login in Settings -> General, switch tabs, confirm state is preserved; verify Open Island appears in System Settings -> General -> Login Items.
- [ ] Manual: turn the toggle off and confirm the entry is removed from Login Items.

## Summary by CodeRabbit

* **New Features**
  * Added a "Launch at login" setting in General preferences so the app can be configured to start automatically at login.
* **Bug Fixes**
  * Improved behavior and rollback when enabling/disabling the setting, and show a user-facing error alert if the system call fails.

### DIFF
--- a/Sources/OpenIslandApp/AppModel.swift
+++ b/Sources/OpenIslandApp/AppModel.swift
@@ -258,6 +258,21 @@ final class AppModel {
             UserDefaults.standard.set(suppressFrontmostNotifications, forKey: Self.suppressFrontmostNotificationsDefaultsKey)
         }
     }
+    var launchAtLoginEnabled: Bool = false {
+        didSet {
+            guard !isApplyingLaunchAtLogin, hasFinishedInit, launchAtLoginEnabled != oldValue else { return }
+            do {
+                try LaunchAtLoginService.shared.setEnabled(launchAtLoginEnabled)
+            } catch {
+                lastActionMessage = "Launch at Login failed: \(error.localizedDescription)"
+                isApplyingLaunchAtLogin = true
+                launchAtLoginEnabled = oldValue
+                isApplyingLaunchAtLogin = false
+            }
+        }
+    }
+    @ObservationIgnored
+    private var isApplyingLaunchAtLogin = false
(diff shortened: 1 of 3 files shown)


Octane0411:
Thanks for the review. Addressed in e731022:

- **[Major] `register()` not idempotent** — added `guard !isEnabled else { return }` on the enable path. Real scenario: if the user toggles login items from System Settings before opening our window, our cached model is stale and a re-register would throw `kSMErrorAlreadyRegistered`.
- **[Major] Errors invisible in Settings** — `lastActionMessage` only renders in the debug ControlCenter card, so a failed toggle silently reverted. Now surfaces via `NSAlert` in the catch block (matches macOS convention for rare modal errors; no new published state needed).
- **[Nitpick] Unused `Logger`** — removed. Was a leftover from an earlier draft.

Skipped:
- **[Nitpick] Comment on init ordering** — every setting in `AppModel.init` follows the same "assign before `hasFinishedInit = true`" pattern (`showDockIcon`, `hapticFeedbackEnabled`, …). Singling this line out would be inconsistent with the rest of the file, and CLAUDE.md defaults to no comments unless the _why_ is non-obvious.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

PR #426, opened 2026-04-29
fix: surface "Open Island hooks not installed" in diagnostics

MVPGFC:
## Summary

The Control Center's **Hooks 诊断 / Hook diagnostics** panel reports "All hooks are healthy" / "所有 Hooks 运行正常" even when **no Open Island hook entries are wired into the agent's config file**. The diagnostic only verifies:

1. ✅ Open Island's own hook *binary* exists at `~/Library/Application Support/OpenIsland/bin/OpenIslandHooks`
2. ✅ `~/.claude/settings.json` (or `~/.codex/hooks.json`) is valid JSON
3. ⓘ Detects third-party hook commands — but flags them only as `.info` severity

It never asks the question: *"is there actually a hook entry that calls our binary?"* So a user with `settings.json` containing only legacy / third-party / forked-product hooks (and no Open Island entry) sees a green checkmark while `Stop`, `SessionEnd`, etc. fire into the void. Sessions appear permanently "running" in the island, which makes the product look broken even though the rest of the integration is fine.

I hit this exact false-positive while testing this dev branch on a machine that previously had a forked product installed: the diagnostic reported all green, but the dev app received zero hook events.

## Fix

`Sources/OpenIslandCore/HookHealthCheck.swift`:

- New `Issue.notInstalled(configPath:)` case
  - `severity = .error`
  - `isAutoRepairable = true` → the existing **Repair** button reinstalls hooks, fixing it in one click
- Emitted from `checkClaude` and `checkCodex` when the agent's config directory exists *and* `hasOpenIslandHooks(...)` returns false
- Suppressed when the config file is malformed JSON (already covered by `.configMalformedJSON`, no need to pile on)
- Suppressed when the agent's config directory doesn't exist (don't nag users who aren't running that agent)

OpenCode already has the analogous `.pluginMissing` issue and is unchanged.

## Test plan

- [x] `swift build` succeeds
- [x] `swift test` — full suite passes (229 tests, 25 suites)
- [x] New `HookHealthCheckTests` (7 tests) covers:
  - third-party-only hooks → `.notInstalled` reported
  - `settings.json` missing but `~/.claude/` exists → `.notInstalled` reported
  - Open Island hooks correctly installed → healthy
  - no `~/.claude/` directory at all → no `.notInstalled` (don't nag)
  - malformed JSON → only `.configMalformedJSON` (no `.notInstalled` pile-on)
  - Codex variant → `.notInstalled` reported when only foreign hooks present
  - `Issue.notInstalled` properties: severity = error, isAutoRepairable = true
- [x] Manual UI verification of the new state in Control Center is covered by existing render paths (`issue.description` is rendered as-is by both `SettingsView.issueList` and `ControlCenterView`); no UI-string code changes required.

## Notes

- No new public API on `HookHealthReport` other than the new enum case.
- Existing call sites that switch over `Issue` will get a compile error if exhaustive — checked: no exhaustive switches over `Issue` exist outside the enum's own `description` / `severity` / `isAutoRepairable` properties, all updated.
- The new check is conservative: it only reports `.notInstalled` when we have positive evidence the agent is in use (its config directory is present), so first-time users who don't run Claude Code / Codex won't see false alarms.

## Summary by CodeRabbit

* **New Features**
  * Improved health diagnostics to report a new "not installed" hook state and offer automatic repair when hooks are missing.
  * Detection refined to avoid reporting missing hooks when configuration files are unreadable or malformed.

* **Tests**
  * Added comprehensive tests covering various installation, unreadable/malformed config, and healthy scenarios.

### DIFF
--- a/Sources/OpenIslandCore/HookHealthCheck.swift
+++ b/Sources/OpenIslandCore/HookHealthCheck.swift
@@ -130,6 +130,7 @@ public enum HookHealthCheck {
         // 2. Check config file
         let settingsPath = settingsURL.path
         var configMalformed = false
+        var configReadFailed = false
         if fileManager.fileExists(atPath: settingsPath) {
             if let data = try? Data(contentsOf: settingsURL) {
                 // Check JSON validity
@@ -155,18 +156,26 @@ public enum HookHealthCheck {
                         issues.append(.otherHooksDetected(names: otherNames.sorted()))
                     }
                 }
+            } else {
+                // File exists but couldn't be read (permissions / IO error).
+                // Don't trust subsequent re-reads of the file, and don't claim
+                // "not installed" when we genuinely don't know what's there.
+                configReadFailed = true
             }
         }
 
         // 3. Detect missing Open Island installation. The presence of the
         // .claude directory is a strong signal that the user runs Claude
         // Code; if our hooks aren't there we should surface it as an error
-        // rather than silently reporting "all healthy".
+        // rather than silently reporting "all healthy". Suppress when the
+        // config couldn't be read or parsed — those failures shouldn't
+        // masquerade as "not installed".
         let claudeDirExists = fileManager.fileExists(atPath: claudeDirectory.path)
-        let openIslandInstalled = !configMalformed
+        let configUntrusted = configMalformed || configReadFailed
+        let openIslandInstalled = !configUntrusted
             && fileManager.fileExists(atPath: settingsPath)
             && hasOpenIslandHooks(in: settingsURL, fileManager: fileManager)
-        if claudeDirExists && !openIslandInstalled && !configMalformed {
+        if claudeDirExists && !openIslandInstalled && !configUntrusted {
             issues.append(.notInstalled(configPath: settingsPath))
         }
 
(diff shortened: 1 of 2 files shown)


MVPGFC:
Thanks for the review @coderabbitai — that's a real edge case I missed. Pushed `ba28824`:

- Track `configReadFailed` separately from `configMalformed`, so a permissions/IO read failure on `settings.json` / `hooks.json` no longer falls through to `.notInstalled`. Both signals gate the new diagnostic via a combined `configUntrusted` flag.
- Applied symmetrically to both `checkClaude` and `checkCodex`.
- Added a regression test (`claudeSkipsNotInstalledWhenSettingsFileUnreadable`) that `chmod 000`s settings.json and asserts no false-positive `.notInstalled`. Self-skips when running as root (where the permission bits don't deny reads).

Kept the existing silent handling of the read failure itself unchanged — surfacing a new `.configUnreadable` issue is a fair follow-up but felt out of scope for this PR. Happy to add it in a separate change if maintainers prefer.

`swift test` passes (230 tests, 25 suites).

PR #433, opened 2026-05-01
debug: add ApprovalFlashDebug logs to trace approval-card auto-dismiss

0xachong:
## Summary
- Adds `os_log` instrumentation under subsystem `app.openisland`, category `ApprovalFlashDebug`, at the three points that together cause the approval/permission card to "弹一下就消失" (flash and hide).
- No behavioral changes — diagnostic only, intended to land before a follow-up fix so we can confirm which path triggers in real Claude/OpenCode/Cursor sessions.

## What's logged
- `BridgeServer.removeClient` → when a hook socket disconnects with a pending Claude/OpenCode/Cursor interaction, before the bridge auto-emits `actionableStateResolved`.
- `SessionState.apply` → when `actionableStateResolved` or `activityUpdated` clears a session's `permissionRequest`.
- `OverlayUICoordinator.reconcileIslandSurfaceAfterStateChange` → when the notification surface no longer matches state and `notchClose()` is forced (with current phase + whether `permissionRequest` is still set).

## Reproduction (verified locally)
A throwaway in-process probe (`OpenIslandFlashProbe`, not included in this PR) drove the bridge end-to-end:
1. Open observer client.
2. Send `processClaudeHook(.permissionRequest)` over a raw socket.
3. Close the socket without waiting for a directive (simulating hook timeout / agent cancel / hook crash).

Observer received: `permissionRequested` → `actionableStateResolved(summary: "Hook process disconnected.")` ⇒ `state.apply` clears `permissionRequest` ⇒ reconcile fires `notchClose()`. Both new os_log warnings fire as expected.

## How to capture in dev
```
/usr/bin/log stream --predicate 'subsystem == "app.openisland" && category == "ApprovalFlashDebug"' --style compact
```

## Test plan
- [ ] Build succeeds (`swift build`) — confirmed locally.
- [ ] Run dev app, trigger a real Claude permission prompt that exhibits the flash, observe `[FLASH]` markers and identify which path fires.
- [ ] No regressions in existing approval flow when the user actually clicks Allow/Deny (logs only fire on the buggy path).

## Summary by CodeRabbit

* **Chores**
  * Enhanced diagnostic logging for notification surfaces and session state transitions.

### DIFF
--- a/Sources/OpenIslandApp/OverlayUICoordinator.swift
+++ b/Sources/OpenIslandApp/OverlayUICoordinator.swift
@@ -2,10 +2,13 @@ import AppKit
 import Foundation
 import Observation
 import OpenIslandCore
+import os
 
 @MainActor
 @Observable
 final class OverlayUICoordinator {
+
+    private static let approvalFlashLogger = Logger(subsystem: "app.openisland", category: "ApprovalFlashDebug")
 
     private static let notificationSurfaceAutoCollapseDelay: TimeInterval = 10
 
@@ -302,6 +305,7 @@ final class OverlayUICoordinator {
             return
         }
 
+        Self.approvalFlashLogger.info("[FLASH] presentNotificationSurface session=\(surface.sessionID ?? "nil", privacy: .public)")
         NotificationSoundService.playNotification(isMuted: isSoundMuted)
         notchOpen(reason: .notification, surface: surface)
     }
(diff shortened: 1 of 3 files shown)


0xachong:
Superseded by #434 — the fix and the diagnostic logs are bundled there because both touch `Sources/OpenIslandCore/BridgeServer.swift` and the project policy forbids chained PRs.

PR #438, opened 2026-05-02
fix(menubar): render brand mark as AppKit template image (#428)

### LINKED ISSUE #428: 显示有个小 bug
Image

Image

不点击的时候是透明的，点击了以后变成一个奇怪的色块了
有人遇到过类似的问题吗？

Karl-Dai:
Closes #428.

## Symptom

Reported by @hetiansu and confirmed by @acewanghaha: the menu-bar status item icon renders incorrectly — nearly invisible while idle, then collapses into a solid black block when the popover is opened. (See screenshots in the issue.)

## Root cause

\`Sources/OpenIslandApp/OpenIslandApp.swift\` uses SwiftUI \`MenuBarExtra\` with \`OpenIslandBrandMark(size: 18, style: .template)\` as its \`label:\`. That label is an arbitrary SwiftUI view containing a \`.drawingGroup\` (which flattens it into an offscreen Metal texture) and \`Color.primary\` for fill.

\`MenuBarExtra\` renders the label, but the result is **not** a real macOS template image (\`NSImage.isTemplate = true\`). Two consequences fall out:

1. While idle, the system can't pick a contrast color → icon looks transparent / a faint block.
2. When the menu-bar item enters the highlighted (selected) state on click, AppKit applies its standard inversion / contrast routine to *non-template* images → the icon collapses into a solid filled block.

## Fix

Pre-rasterize the same 8×8 brand pattern (which is data-driven by \`scoutPattern\`) to an \`NSImage\` with \`isTemplate = true\` and use \`Image(nsImage:)\` as the \`MenuBarExtra\` label. AppKit then handles light / dark / highlight states correctly.

\`OpenIslandBrandMark\`'s SwiftUI \`body\` is **unchanged** — every other use site (island surface, settings panes, onboarding) keeps the duotone SwiftUI path.

## Files

- \`Sources/OpenIslandApp/OpenIslandBrandMark.swift\` (+29) — adds \`makeMenuBarTemplateImage\` and a cached \`menuBarTemplateImage\`. Reuses the existing \`pixels\` data, no duplication of the 8×8 pattern.
- \`Sources/OpenIslandApp/OpenIslandApp.swift\` (~1) — swaps the \`MenuBarExtra\` \`label:\` to \`Image(nsImage: OpenIslandBrandMark.menuBarTemplateImage)\`.

\`+30 / -1\`.

## Test plan

\`swift build\` passes. AppKit rendering can't be unit-tested usefully — manual verification:

- [ ] Light mode → icon is dark, recognizable as the brand pattern.
- [ ] Dark mode → icon auto-inverts to light, still recognizable.
- [ ] Click to open popover → menu-bar item enters highlighted state → icon stays clean and inverted, no longer collapses into a black block.
- [ ] Close popover → icon returns to idle look.
- [ ] Confirm island / settings / onboarding still show the colored duotone brand mark unchanged.

If you'd like, I can attach before/after screenshots of the three states once you've confirmed the approach.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS menu bar icon rendering and visual consistency by replacing the previous rendering with a template-based menu-bar image, resulting in crisper, correctly inverted pixel rendering and consistent opacity handling.

### DIFF
--- a/Sources/OpenIslandApp/OpenIslandApp.swift
+++ b/Sources/OpenIslandApp/OpenIslandApp.swift
@@ -126,7 +126,7 @@ struct OpenIslandApp: App {
         MenuBarExtra {
             MenuBarContentView(model: appDelegate.model)
         } label: {
-            OpenIslandBrandMark(size: 18, style: .template)
+            Image(nsImage: OpenIslandBrandMark.menuBarTemplateImage)
                 .accessibilityLabel("Open Island")
                 .background(SettingsOpenerRegistrar(model: appDelegate.model))
         }
--- a/Sources/OpenIslandApp/OpenIslandBrandMark.swift
+++ b/Sources/OpenIslandApp/OpenIslandBrandMark.swift
@@ -83,7 +83,7 @@ struct OpenIslandBrandMark: View {
         let image = NSImage(size: size, flipped: false) { _ in
             guard let ctx = NSGraphicsContext.current?.cgContext else { return false }
             for pixel in pixels {
-                let alpha: CGFloat = pixel.role == "E" ? 0.7 : 1.0
+                let alpha: CGFloat = pixel.role == "E" ? 0.9 : 1.0
                 ctx.setFillColor(NSColor.black.withAlphaComponent(alpha).cgColor)
                 let rect = CGRect(
                     x: padding + CGFloat(pixel.x) * pixelSize,
--- a/Sources/OpenIslandApp/OpenIslandBrandMark.swift
+++ b/Sources/OpenIslandApp/OpenIslandBrandMark.swift
@@ -1,3 +1,4 @@
+import AppKit
 import SwiftUI
 
 struct OpenIslandBrandMark: View {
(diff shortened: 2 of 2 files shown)


Karl-Dai:
Hi @Octane0411 — friendly nudge for review on this small bug fix when you have a moment.

### TL;DR
The menu-bar icon is currently broken (#428): nearly invisible at idle, collapses to a solid black block when the popover opens. This PR replaces the SwiftUI label with a real AppKit template `NSImage` so macOS can do its standard light / dark / highlight handling.

### Why this is the right fix (and not the alternatives)
- **Just removing `.drawingGroup`** still leaves the label as an arbitrary SwiftUI view. `MenuBarExtra` does not promote such labels to true template images, so the highlight bug stays.
- **Switching to an SF Symbol** would lose brand identity.
- **Pre-rasterizing the same 8×8 pattern to an `NSImage` with `isTemplate = true`** is the standard pattern used by Bartender / Fantastical / Things and matches what `NSStatusItem` expects.

### Scope discipline
- `OpenIslandBrandMark.body` is **unchanged** — island, settings, onboarding all keep the SwiftUI duotone path.
- The 8×8 pattern is reused from the existing `pixels` static; no duplication.
- One new method + one cached `static let`; one line changed in `OpenIslandApp.swift`. `+30 / -1`.

### What I'd appreciate your eyes on
1. The pixel sizing (`pixelSize: 2`, `padding: 1`, total 18×18 points) — happy to adjust if you'd prefer a different visual weight.
2. Whether you want me to also attach before/after screenshots of light / dark / highlight states to the PR description.

Thanks for taking a look — the fix is small and isolated, and the issue has another user reporting the same thing so this should unblock the next release.

Karl-Dai:
Addressed @coderabbitai's note in `6f74cc4` — aligned the E-pixel alpha to `0.9` to match the SwiftUI `.template` branch. The center cell now renders identically across the menu-bar (`NSImage` path) and the in-app uses (SwiftUI `.template`).

PR #439, opened 2026-05-02
feat: ambient theme, celebrations, context bar + cmux/1M-context bug fixes

h4ckm1n-dev:
## Summary

Adds an ambient project theme, celebration confetti, project colors, and a per-session context-usage bar that finally reads correctly on 1M-context Claude Opus sessions. Also fixes several agent / terminal bugs around cmux integration and a transcript watcher crash.

Open Island with expanded notch and per-session context bars

### Bug fixes
- **Context-usage bar now works on 1M Opus sessions.** Claude Code's JSONL doesn't tag 1M-context sessions in the `model` field, so the bar previously fell back to the 160K window when `cache_read` peaked around 175–185K, clamped to 100% used, and rendered as a red dot. Detection threshold lowered from `>200K` to `>165K` (anything past Claude Code's 200K auto-compact threshold without compaction is effectively impossible on a 200K session). Also tracks max-observed used across the whole transcript so a post-compact session that's mid-conversation doesn't briefly fall back to the wrong window.
- **cmux tab-switching from the notch.** Captures `CMUX_SURFACE_ID` from the agent process env (`ps eww`) into the `JumpTarget` and sends `surface.focus` over cmux's Unix socket, so clicking a session row jumps to the right cmux tab.
- **Auto-detect open cmux tabs** for sessions that previously rendered as "Unknown" terminal — matches Unknown-terminal sessions to live cmux processes by `cwd`.
- **Reply-to-session works for cmux.** `tmuxTarget` resolution now also runs for cmux sessions, so the completion-reply UX in the expanded view actually sends keystrokes back to the right tab.
- **DispatchSource @MainActor crash** in the transcript watcher fixed via `WeakBox`-wrapped `@Sendable` event handlers.
- **`launch-dev-app.sh`** prefers an existing `Apple Development` cert over the local self-signed one — avoids macOS 26 SIGKILL/CODESIGNING rejection on dev builds.

### Personalization (all opt-in via Settings → Appearance)
- **Ambient project theme** — subtle linear-gradient overlay tinted by the spotlight project's color (FNV-1a hash of the workspace key, with overrides). Toggle + opacity slider.
- **Celebration confetti** — 12-particle deterministic-physics confetti (Canvas) that spawns on session completion. Respects Reduce Motion.
- **Project colors settings subsection** — disclosure group lists known workspace keys with a color picker and a reset-all button.

### Context bar
- **Per-session `ContextLeftBadge`** in expanded session rows — a 4×18 fill bar plus percent label, color-graded green/yellow/orange/red based on remaining context. The label shows **percent used** (so a near-full bar reads as a high number).

## Test plan

- [x] `swift build` clean
- [x] `swift test` — **272/272 passing.** New regression coverage: ContextLeftBadge fills/colors, ContextWindowTable thresholds, ContextUsageReader 1M detection at the 180K cache_read boundary, post-compact 1M lock, ProjectColorRegistry persistence, CelebrationParticles math, AmbientThemeOverlay opacity.
- [x] Dev build verified on macOS 14+ with Claude Code, Codex, and cmux sessions: ambient gradient applies, celebration fires on completion, context bar reads correctly on a real 1M-Opus session at ~193K used (~24% of 800K), cmux tab-switch jumps to the right tab.
- [x] No regressions on Claude / Codex / OpenCode session reconciliation, transcript discovery, or bridge protocol.

## Notes for review

- This PR is intentionally scoped to features and fixes that don't require upstream config changes. Happy to split into smaller PRs (bug fixes vs personalization vs context bar) if that's preferred — let me know.
- A bilingual (EN / 中文) changelog can be added before release if needed; flagging because the project's `RELEASE_TEMPLATE.md` requires it for tagged releases.

## Summary by CodeRabbit

* **New Features**
  * Ambient theme with adjustable opacity and project-color personalization (reset/prune)
  * Celebration animations and companion state overlay
  * Context usage indicators and workspace-aware project colorization

* **Improvements**
  * More robust terminal/session matching including better tmux/cmux attribution and session identifiers

* **Localization**
  * Added Simplified and Traditional Chinese entries (English placeholders pending translation)

* **Tests**
  * Extensive unit tests for visuals, context usage, and project color registry


h4ckm1n-dev:
Thanks for the review. Pushed `c836457` addressing the actionable items:

**Addressed**

- **`ActiveAgentProcessDiscovery` cmux/Unknown-host gap.** `CMUX_SURFACE_ID` is now read first and used as authoritative cmux evidence — its presence labels `terminalApp = "cmux"` when discovery couldn't, so subsequent tmux pane resolution and the surface-ID-based jump both populate even for "Unknown" hosts. Collapsed the three previously-duplicated inline blocks (codex / claude / opencode) into a single `resolveCmuxAndTmuxMetadata` helper.
- **`ContextUsage` + `CompanionState` Codable.** Added per the `OpenIslandCore` "Sendable + Codable" guideline. Auto-synthesis works on both.
- **`CompanionStateOverlay` accessibility labels.** Now route through `LanguageManager` with new `island.companion.state.{idle,working,waiting,celebrating}` keys in en / zh-Hans / zh-Hant.

**Already addressed by prior cleanup commits on this branch**

- The `docs/plans/2026-04-30-pets-ambient-celebrations-design.md` and `docs/plans/2026-05-02-remove-notch-widgets.md` flagged as outdated were dev-only artifacts and were removed in `bbf1a29`. The pet feature itself was removed in `b09ccb1`, so the 8-pet manifest discussion no longer applies.
- `CodeburnClient` / `CodeburnTypes` Codable concerns are moot — both files were removed in `bbf1a29` along with the rest of the codeburn integration.

**Not addressed (deferred — happy to follow up if you'd like a separate fix)**

- *"Time-window celebration state can persist longer than intended"* in `IslandPanelView`. The 8-second window relies on SwiftUI re-evaluation; in practice the spotlight session and other observed properties change frequently enough that the visible artifact is small, and `CelebrationParticles` is `id`'d on the timestamp so it cleans itself up. I'd rather wrap this in a `TimelineView` or explicit timer in a follow-up so it can ship with a regression test, instead of bundling it into this already-large PR.

`swift test` 272/272 still green.

h4ckm1n-dev:
Pulled the deferred fix into this PR — `89a0915`.

**What changed**

- New `CompanionState.celebratingWindow` constant + pure-function helper `isWithinCelebratingWindow(now:lastCompletion:window:)` in `OpenIslandCore`.
- `IslandPanelView`'s `companionState` computed property now uses the helper instead of an inline `Date()` comparison.
- Two new `.task(id:)` modifiers attached to `lastCompletionTimestamp` and `lastCelebrationTimestamp`: they sleep until the relevant window elapses, then nil out the binding. The state mutation guarantees a SwiftUI re-render at the boundary, so the celebrating glyph and the particle container both clean up on schedule. Each new timestamp cancels the prior pending expiry via Task identity.

**Tests**

- 6 new boundary tests in `CompanionStateTests`: nil completion, zero elapsed, just-before / at / after the boundary, and a custom-window case.
- `swift test` — 278/278 passing.

h4ckm1n-dev:
Verified and fixed in `3cd1adb`.

**Verified the bug:** `@State` persists across re-mounts of the same view identity. The `.working` branch's `onAppear` calls `animateRotation()` which writes `rotation = 360` inside `withAnimation(.repeatForever)`. After the first activation, `rotation` is `360` — when the state cycles `working → idle → working` and `onAppear` fires again, the write is to the same value, SwiftUI sees no delta, and the implicit animation never restarts. Same shape for `.waiting`'s pulse.

**The fix:** reset-before-animate.

```swift
private func animateRotation() {
    var resetTransaction = Transaction()
    resetTransaction.disablesAnimations = true
    withTransaction(resetTransaction) { rotation = 0 }
    DispatchQueue.main.async {
        withAnimation(.linear(duration: 2).repeatForever(autoreverses: false)) {
            rotation = 360
        }
    }
}
```

The transactionless reset snaps `rotation` back to `0` without animation; the `DispatchQueue.main.async` defers the `withAnimation` block to the next runloop tick so SwiftUI commits the reset first and then sees an actual delta when the animation kicks off. Same pattern applied to `animatePulse`.

`swift test` 278/278 still green.

h4ckm1n-dev:
Friendly nudge — would be great to get this merged when you have time. There's a sizable follow-up branch (`feat/inline-diff-and-plan-mode` on my fork) sitting on top of this one, and per the project's chain-PR rule it can't be opened until this one lands on `main`.

Here's what's ready in the follow-up so you can gauge whether it's worth prioritizing this merge:

### Killer features

- **Inline diff preview in the approval card** — when Claude Code requests permission to mutate a file (`Edit` / `Write` / `MultiEdit`) or Codex requests `apply_patch`, the approval card renders a unified diff with red/green tinted rows + lightweight syntax highlighting (Swift, TS, JS, Python, Rust, Go, Ruby, shell). Lets you decide Yes/No without switching back to the terminal. Backed by an O(n·m) Myers diff capped at 2000 lines per side.
- **Plan mode visualization** — when an agent submits a plan (native `ExitPlanMode`, or markdown written via the `superpowers:brainstorming` / `superpowers:writing-plans` skills, or GSD `.planning/`), the notch parses it into a checklist. Pre-approval shows a read-only review block; post-approval the row gains a `📋 Plan (n/m)` disclosure with interactive checkboxes that survive quit/relaunch via UserDefaults.

### Security hardening

- **Trojan Source / zero-width Unicode protection** — bidi overrides (U+202A–U+202E, U+2066–U+2069), zero-width chars (U+200B–U+200F, U+00AD, U+FEFF), and U+2028/U+2029 separators in diff content are replaced with visible `⟨U+XXXX⟩` placeholders + an `eye.trianglebadge.exclamationmark` row badge so a malicious agent can't display one thing while writing another.
- **File-path display hardening** — full `file_path` shown with middle truncation (no more hiding `/Users/x/.ssh/authorized_keys` behind a benign `.ssh/...` tail). Yellow caption when the path traverses with `..` or writes outside `$HOME`.
- **Multi-file `apply_patch` rejection** — patches with more than one `+++ b/...` header now refuse to render preview (with a "review in terminal" stub) so the agent can't bury a sensitive write under a benign-looking last filename.
- **Documented audit** under `docs/security/audit-2026-05-03.md` — full P0 report covering the threat model, plus catalogued findings F5/F6/F7 deferred for future PRs.

### Theme system

- **Catppuccin theme picker** in Settings → Appearance — 4 official Catppuccin flavors (Latte / Frappé / Macchiato / Mocha) plus a `System` option that preserves the pre-theme look. Live 8-swatch preview row updates as you change selection.
- **Centralized `ThemePalette`** in `OpenIslandCore` with semantic Catppuccin tokens (base, mantle, crust, surface0–2, text, subtext0–1, overlay0–2, plus 14 accents). Inline diff colors, syntax highlighter palette, agent/terminal badge tints, status indicator colors, companion-state overlay glyphs, context bar thresholds, and project color presets all read from the active palette and repaint instantly on theme switch via `@Observable`.

### Bug fixes carried in the follow-up

- **`TerminalTextSender` runs AppleScript on the main thread** — fixes silent no-op when reply submission was dispatched on `Task.detached`.
- **cmux native (no tmux backend) reply support** — `sendViaCmux` focuses the surface via the existing JSON-RPC socket and drives a System Events keystroke under the cmux process.
- **`CompanionStateOverlay` animation restart** — reset-before-animate pattern fixes the rotation/pulse freezing on the second visit to `.working` / `.waiting` after `@State` retained the target value.

---

**Stats:** 14 new commits, 6 new features, 7 bug fixes, ~50 new tests. `swift build` clean and `swift test` 355/355 passing on the follow-up branch.

If you'd prefer this PR be split into smaller chunks before merging, happy to do that — just say the word. Otherwise, no rush, but flagging because the follow-up work is shippable and currently parked.

PR #463, opened 2026-05-09
fix(codex): update hooks feature flag (issues #460)

### LINKED ISSUE #460: codex (v0.130.0)   `[features].codex_hooks` is deprecated. Use `[features].hooks` instead.
Tip: New Build faster with the Codex App. Run 'codex app' or visit https://chatgpt.com/codex?app-landing-page=true

⚠ `[features].codex_hooks` is deprecated. Use `[features].hooks` instead.
Enable it with `--enable hooks` or `[features].hooks` in config.toml. See https://developers.openai.com/codex/config-basic#feature-flags for detail
s.

⚠ 3 hooks need review before they can run. Open /hooks to review them.

ViperThanks:
Fixes #460.

Codex v0.130 deprecated `[features].codex_hooks` in favor of `[features].hooks`.

This PR updates the managed Codex hook installer to use the current feature flag while keeping existing installs compatible.

Changes:

- write `[features].hooks = true` for new managed installs
- treat both `hooks = true` and legacy `codex_hooks = true` as enabled during status checks
- migrate legacy `codex_hooks = true` to `hooks = true` on managed install
- remove stale `codex_hooks = true` when both current and legacy flags are present, so Codex v0.130 no longer prints the deprecation warning
- keep uninstall conservative: only remove the feature flag when the manifest says Open Island enabled it
- document Codex's `/hooks` review step

Compatibility note:

This intentionally writes only the current `hooks = true` flag. Keeping `codex_hooks = true` would preserve compatibility with older Codex builds but would continue to trigger the v0.130 deprecation warning reported in this issue.

Existing installs are still recognized before migration because status checks accept both `hooks = true` and legacy `codex_hooks = true`.

The second warning in the issue:

> 3 hooks need review before they can run. Open /hooks to review them.

is Codex CLI's own trust gate. Open Island should not bypass it; users still need to approve the expected hooks in `/hooks`.

Validation:

- `git diff --check`
- `scripts/check-docs.sh`
- `swift build --product OpenIslandHooks`
- `swift build --product OpenIslandSetup`

Note: `swift test --filter 'SessionStateTests/codexHookInstaller'` is blocked in my local environment because the existing test target cannot import Swift's `Testing` module. The product builds pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable Codex hooks detection and management: legacy hook settings are migrated to the modern flag and the current hooks flag is ensured/removed as appropriate.

* **New Features**
  * Added explicit feature-detection logic so installs consistently recognize enabled hooks (including legacy variants).

* **Tests**
  * Expanded tests verifying migration, removal, and detection behavior.

* **Documentation**
  * Updated installation docs for v0.130+ and noted manual trust approval may still be required.

### DIFF
--- a/Sources/OpenIslandCore/CodexHookInstallationManager.swift
+++ b/Sources/OpenIslandCore/CodexHookInstallationManager.swift
@@ -67,7 +67,7 @@ public final class CodexHookInstallationManager: @unchecked Sendable {
             hooksURL: hooksURL,
             manifestURL: manifestURL,
             hooksBinaryURL: resolvedHooksBinaryURL,
-            featureFlagEnabled: configContents.contains("codex_hooks = true"),
+            featureFlagEnabled: CodexHookInstaller.isCodexHooksFeatureEnabled(in: configContents),
             managedHooksPresent: managedHooksPresent,
             manifest: manifest
         )
--- a/Sources/OpenIslandCore/CodexHookInstaller.swift
+++ b/Sources/OpenIslandCore/CodexHookInstaller.swift
@@ -59,6 +59,8 @@ public enum CodexHookInstaller {
     public static let managedStatusMessage = "Managed by Open Island"
     public static let legacyManagedStatusMessage = "Managed by Vibe Island"
     public static let managedTimeout = 45
+    private static let currentFeatureKey = "hooks"
+    private static let legacyFeatureKey = "codex_hooks"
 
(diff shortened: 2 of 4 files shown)


ViperThanks:
Fixed in 77bee8e.

The feature flag parser now accepts TOML-style variants such as `hooks=true` and `hooks = true # comment`, and the installer no longer relies on exact string equality for ownership inference. I also added regression coverage for compact assignment and inline-comment forms.

PR #489, opened 2026-05-13
[codex] fix Claude subagent stop completion regression

Octane0411:
## Summary
- Keep completed Claude sessions completed when a late `SubagentStop` hook arrives after `Stop`.
- Prevent subagent lifecycle `agent_id` / `agent_type` values from being persisted onto the parent Claude session metadata.
- Add a regression test for `UserPromptSubmit -> Stop -> SubagentStop(agent_id, agent_type: "")`.

## Root Cause
Claude Code can emit a normal `Stop` hook for the parent turn and then emit a later `SubagentStop` payload with an `agent_id` and empty `agent_type`. The bridge previously handled every `SubagentStop` by emitting `activityUpdated(... phase: .running)`, which could reopen an already completed parent session and leave it stuck as running.

## Validation
- `swift test --filter ClaudeHooksTests`
- `swift test` before rebasing this branch
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented reopening or further updates to a session after it has completed when a subagent stops.
  * Avoided emitting redundant metadata updates and preserved subagent identity during lifecycle changes.

* **Tests**
  * Added an integration test to validate session remains completed when a subagent stops, and a safe async event collector for the test.

### DIFF
--- a/Sources/OpenIslandCore/BridgeServer.swift
+++ b/Sources/OpenIslandCore/BridgeServer.swift
@@ -892,10 +892,18 @@ public final class BridgeServer: @unchecked Sendable {
         case .subagentStop:
             ensureClaudeSessionExists(for: payload)
             synchronizeClaudeJumpTarget(for: payload)
-            synchronizeClaudeMetadata(for: payload)
+            let sessionWasAlreadyCompleted = localState.session(id: payload.sessionID)?.phase == .completed
+            if !sessionWasAlreadyCompleted {
+                synchronizeClaudeMetadata(for: payload)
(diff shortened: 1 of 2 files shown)


Octane0411:
Updated this branch to current `main` and addressed the review feedback by replacing the fixed sleep in the regression test with a bounded wait for the observer to receive the completed-session event.

Validation:
- `git diff --check`
- `swift build`

`swift test` remains blocked in this checkout by the local test compile error `no such module 'Testing'`.

PR #496, opened 2026-05-16
fix: strip markdown syntax from preview text and completion card

Ethanlam123:
## Summary

- Replace `MarkdownUI` rendering in the completion card with plain `Text`, stripping markdown syntax before display
- Strip markdown from all surface preview text (`trimmedForSurface`) so island spotlight shows clean text instead of raw markdown markers
- Two stripping variants: `strippingMarkdown` (collapses whitespace, for single-line previews) and `strippingMarkdownForCard` (preserves newlines, for card body)

Closes #58

## Changes

- `AgentSession+Presentation.swift`: `trimmedForSurface` now calls `strippingMarkdown` before trimming, removing code blocks, inline code, bold/italic, headings, links, and list markers
- `IslandPanelView.swift`: Completion card body uses `Text(completionMessageText.strippingMarkdownForCard)` instead of `Markdown(completionMessageText).markdownTheme(.completionCard)`

## Test plan

- [x] `swift build` passes
- [x] `swift test` - no regressions (5 pre-existing failures on base branch)
- [ ] Visual verification via harness: `OPEN_ISLAND_HARNESS_SCENARIO=longCompletionCard OPEN_ISLAND_HARNESS_PRESENT_OVERLAY=1 swift run OpenIslandApp`
- [ ] Manual test with live agent session via dev app

## Summary by CodeRabbit

* **Bug Fixes**
  * Session spotlight text now has Markdown formatting removed before display, preventing raw markup from appearing in UI.
  * Completion messages are shown as sanitized plain text (Markdown stripped) so cards render cleanly and consistently without formatting artifacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Octane0411/open-vibe-island/pull/496?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/OpenIslandApp/AgentSession+Presentation.swift
+++ b/Sources/OpenIslandApp/AgentSession+Presentation.swift
@@ -442,6 +442,19 @@ extension AgentSession {
 
 private extension String {
     var trimmedForSurface: String {
-        trimmingCharacters(in: .whitespacesAndNewlines)
+        strippingMarkdown.trimmingCharacters(in: .whitespacesAndNewlines)
+    }
+
+    var strippingMarkdown: String {
+        var s = self
+        s = s.replacingOccurrences(of: "```[\\s\\S]*?```", with: "", options: .regularExpression)
+        s = s.replacingOccurrences(of: "`([^`]+)`", with: "$1", options: .regularExpression)
+        s = s.replacingOccurrences(of: "\\*{1,3}(.+?)\\*{1,3}", with: "$1", options: .regularExpression)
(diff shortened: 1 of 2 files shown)


Ethanlam123:
Addressed in `178cf94`:

- **Multiline flag fix** (both files): Added `(?m)` to the list/quote marker pattern `^[\-*>+]\s+` so it matches at the start of every line, not just the first.
- **Extract shared helper** (nitpick): Intentionally skipped. The two `private extension String` blocks are file-scoped and differ only in the final whitespace collapse step. The duplication is ~10 lines and doesn't justify cross-file indirection. If a third consumer appears, that would be the time to consolidate.

PR #540, opened 2026-06-05
feat: add Pi extension support

caiguo37:
## Summary
- add Pi extension hook parsing and bridge handling
- add bundled Pi extension resource and install/status management
- surface Pi extension setup in settings and docs

## Verification
- `swift test --filter PiHooksTests` built successfully, but local test launch failed because this machine could not load `libsqlite3-1.dylib` from the Xcode/macOS runtime search paths.

## Summary by CodeRabbit

* **New Features**
  * Added Pi as a supported agent with UI integration, install/uninstall controls, and tool-labeling ("Pi")
  * Desktop bridge integration to connect Pi sessions to the app via a shipped extension and Unix socket
  * Pi tool approval/deny workflow for blocking tool calls

* **Bug Fixes**
  * Improved Pi session liveness and heartbeat handling to prevent premature session termination

* **Documentation**
  * Docs updated with Pi architecture, install instructions, event mapping, and directive format

* **Tests**
  * Added tests validating Pi hook payloads and approval flow

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Think of it as an open-source [Vibe Island](https://vibeisland.app/) — **free,
 
 ## Supported Agents & Terminals
 
-**10 agents**: Claude Code, Codex, Cursor, Gemini CLI, Kimi CLI, OpenCode, Qoder, Qwen Code, Factory, CodeBuddy
+**11 agents**: Claude Code, Codex, Cursor, Gemini CLI, Kimi CLI, OpenCode, Pi, Qoder, Qwen Code, Factory, CodeBuddy
 
 **15+ terminals & IDEs**: Terminal.app, Ghostty, iTerm2, WezTerm, Zellij, tmux, cmux, Kaku, VS Code, Cursor, Windsurf, Trae, JetBrains IDEs (IDEA, WebStorm, PyCharm, GoLand, CLion, RubyMine, PhpStorm, Rider, RustRover)
 
@@ -72,6 +72,7 @@ Think of it as an open-source [Vibe Island](https://vibeisland.app/) — **free,
 | **Cursor** | Supported | Hook integration via `~/.cursor/hooks.json`, session tracking, workspace jump-back |
 | **Gemini CLI** | Supported | Hook integration via `~/.gemini/settings.json`, session tracking, fire-and-forget events |
 | **Kimi CLI** | Supported | Hook integration via `~/.kimi/config.toml` `[[hooks]]`, session tracking, permission flow (reuses Claude payload) |
+| **Pi** | Supported | Pi Extension API bridge, session tracking, tool approval for mutating built-in tools |
 
 ### Terminals & IDEs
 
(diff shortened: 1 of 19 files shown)


caiguo37:
A couple of design notes for the Pi integration:

1. **Why Pi sessions are heartbeat-managed**

Pi runs through the Extension API rather than a CLI hook process that gives Open Island a stable app-side process/session identifier. Because of that, the generic process polling fallback cannot reliably match a Pi session to a local process. If we let that fallback apply unchanged, Pi rows can be marked ended in the gap between extension events.

That is why Pi uses an extension heartbeat as the liveness signal:

- `Heartbeat` keeps/restores the session while Pi is alive.
- Generic process polling does not end Pi sessions.
- `SessionEnd` from Pi’s `session_shutdown` still closes the session when Pi actually shuts down.

This keeps Pi lifecycle source-specific and avoids guessing from unrelated local process state.

2. **Why approval behavior is intentionally limited**

Pi does not expose the same first-class approval protocol as Codex/Claude hooks. The Extension API lets us intercept tool calls and block them, but it does not provide a stable contract for richer permission-mode changes, persistent allow rules, or tool-input rewriting.

So Open Island only brokers approval for the mutating built-in tools we can safely identify (`bash`, `edit`, `write`):

- allow = return normally and let Pi continue
- deny = return Pi’s blocking extension response

We intentionally do not patch Pi internals, bypass Pi runtime behavior, or invent broader approval semantics that Pi itself does not expose through the public Extension API.

lalitjeevanrao:
Hello @caiguo37 , do you have a roadmap to review this PR and merge? I would like to use this for my PI coding agent.
Thanks in advance!

mswell:
I tested the Pi + Herdr integration on macOS 26.5.2 with Open Island.

The fixes are available in my fork:
https://github.com/mswell/open-vibe-island/tree/fix/pi-herdr-jump

These commits apply on top of this PR:

- 99e6cf4: focus Pi notifications in Herdr panes
- dca9604: preserve Herdr Pi jump targets
- 8a21b48: request macOS Accessibility permission
- e247733: use Herdr pane identity from the Pi environment

They address:

- Pi targets being reclassified from Herdr to Ghostty by the attachment reconciler.
- Focusing Herdr itself without selecting the originating pane.
- Pane identity becoming stale when Herdr recreates panes.
- Open Island not being registered in macOS Accessibility automatically.

Manual end-to-end test passed: reload the Pi extension, start a task in one Herdr pane, switch to another pane, wait 20-30 seconds, and click the Open Island notification. The exact originating pane is focused.

The original PR is owned by caiguo37, so I could not push these commits directly to its branch. Could you cherry-pick them, or let me know if you prefer a follow-up PR?

mswell:
Follow-up: the tested local fixes are now pushed to the same branch.

New commit: `18afbb3 fix: recover local agent sessions in Herdr`

It adds Claude Code recovery when Open Island starts after Claude, maps Claude Herdr sessions to their pane IDs, adds the option to disable Open Island Pi tool approvals, and removes the unnecessary automatic Accessibility prompt for Pi/Herdr.

Octane0411:
Thanks for this, and for the early testers it attracted. Pi support has now landed through #639 (Pi + Oh My Pi, extension-based, heartbeat liveness, process discovery), which covers everything this PR does except the blocking approvals — and those were left out deliberately: Pi ships without a permission system by design, so turning every `bash`/`edit`/`write` into an Open Island approval would add a gate the agent never had (one tester already needed a switch to turn it off). If you'd like approvals as an opt-in setting on top of the merged integration, a follow-up PR using #639's `tool_call` path would be very welcome. Closing this one as superseded.

_This comment was generated by AI and may contain mistakes in judgement. If anything here is wrong or unclear, please reply directly to this comment._

PR #541, opened 2026-06-05
feat: add menu bar status item mode

caiguo37:
## Summary
- add an optional menu bar status item mode as an alternative to the notch/top-bar overlay
- render live session state chips in the menu bar and anchor the existing island panel below the status item
- grey out notch-only settings while menu bar mode is enabled and avoid recursive settings reopening

## Verification
- `swift build`

## Summary by CodeRabbit

* **New Features**
  * Menu bar status item showing session counts/status with color indicators and toggling the overlay
  * Settings toggle to enable/disable the menu bar icon; settings UI shows a note and disables conflicted options when enabled
  * Menu-bar mode: compact header, adjusted panel positioning and sizing for menu bar placement

* **Bug Fixes**
  * Settings window reliably brought to the front after opening

* **Localization**
  * Added English, Simplified Chinese, and Traditional Chinese strings

### DIFF
--- a/Sources/OpenIslandApp/AppModel.swift
+++ b/Sources/OpenIslandApp/AppModel.swift
@@ -17,6 +17,7 @@ extension Notification.Name {
 final class AppModel {
     private static let soundMutedDefaultsKey = "overlay.sound.muted"
     private static let showDockIconDefaultsKey = "app.showDockIcon"
+    private static let menuBarStatusItemDefaultsKey = "app.menuBarStatusItem"
     private static let hapticFeedbackEnabledDefaultsKey = "app.hapticFeedbackEnabled"
     private static let islandRightSlotDefaultsKey = "appearance.island.v6.rightSlot"
     private static let islandCenterLabelDefaultsKey = "appearance.island.v6.centerLabel"
@@ -49,6 +50,7 @@ final class AppModel {
             _cachedSessionBuckets = nil
             pruneAgentsGridObservationTicketsIfNeeded()
             bridgeServer.updateStateSnapshot(state)
+            statusItemController?.refreshStatus()
         }
     }
     @ObservationIgnored private var _cachedSessionBuckets: (primary: [AgentSession], overflow: [AgentSession])?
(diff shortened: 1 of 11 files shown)


caiguo37 (commented):
Menu bar status item screenshot

Design notes for this PR:

### Why this should not break the existing notch presentation

The menu bar status item path is intentionally feature-gated behind the new `menuBarStatusItemEnabled` preference. The existing notch/top-bar presentation remains the default because:

- `menuBarStatusItemEnabled` defaults to `false` unless the user explicitly enables it in Settings.
- `OverlayPanelController.menuBarMode` also defaults to `false`, so the original hover, click, closed-pill, and screen-centered placement paths are unchanged by default.
- All menu-bar-specific behavior is guarded by `menuBarMode` / `model.menuBarStatusItemEnabled`, including:
  - hiding the floating closed pill,
  - disabling hover-open for the closed surface,
  - anchoring the opened panel below the `NSStatusItem`,
  - forcing a flat non-notch surface, and
  - using the measured menu-bar content height.
- When menu bar mode is disabled, panel placement still uses the original notch/top-bar frame calculation centered at the top of the screen, and the original closed-surface rendering remains active.

So this PR adds an alternate presentation mode rather than replacing the notch path. Toggling the setting off removes the status item, clears the menu-bar anchor providers, and returns the overlay controller to the normal notch/top-bar behavior.

### Main design approach

The goal is to support users who prefer a standard macOS menu bar indicator while reusing the existing island UI stack instead of creating a second UI implementation.

Key choices:

- Add a lightweight `StatusItemController` that owns the `NSStatusItem`, renders compact per-state session chips, and toggles the existing island panel.
- Reuse `IslandPanelView` and `OverlayPanelController` for the opened dropdown so session rows, actions, settings, sound controls, and quit behavior stay consistent with the notch UI.
- Treat the status item as the closed state in menu bar mode. The floating closed pill is hidden only in this mode to avoid two simultaneous closed indicators.
- Anchor the opened panel under the status item button and clamp it on-screen, matching normal macOS menu bar dropdown expectations.
- Move the session overview into the top control strip in menu bar mode, removing the duplicate standalone header and keeping the dropdown compact.
- Disable or explain notch-only settings while menu bar mode is enabled, because auto-collapse, monitor/position, and closed-pill appearance settings do not apply to the menu bar presentation.

Verification performed locally: `swift build`.

PR #578, opened 2026-07-06
fix: age completed sessions out of island primary

### LINKED ISSUE #576: Completed Claude sessions remain counted as idle while daemon is alive
## Environment

- Open Island: local 1.1.4 build, reproduced after merging the latest `origin/main`
- macOS: notch MacBook
- Terminal / IDE involved: Claude Code launched from VS Code / Claude daemon background process
- Session registry involved: `~/Library/Application Support/open-island/claude-session-registry.json`

## Description

Completed Claude Code sessions can remain visible in the expanded island under Idle and keep contributing to the total / idle counters long after the work is done, if the Claude daemon or background process remains alive.

In the observed state, the island showed several old idle Claude rows even though only the current sessions were actionable. Some restored rows had `terminalApp = Unknown` because the saved jump target no longer had a terminal app, but the main problem was that stale completed sessions were still being treated as primary visible sessions.

This is similar in shape to #542, but this report is for Claude sessions whose completed rows are protected by Claude daemon / background process liveness instead of Codex.app liveness.

## Steps to Reproduce

1. Start one or more Claude Code sessions from VS Code / Claude Code so the Claude session registry records them.
2. Let those sessions finish or reach a completed / waiting-for-input state.
3. Keep the Claude daemon or background process alive.
4. Wait beyond the completed stale threshold.
5. Open the expanded island session list with grouping enabled.

## Expected vs Actual Behavior

Expected:

- Old completed Claude sessions should age out of the primary island list after the configured completed-stale threshold.
- Those sessions can remain available in recent / overflow history if needed.
- Header counts should reflect only currently relevant primary sessions.

Actual:

- Completed Claude rows can remain visible in the Idle section for a long time.
- They keep inflating the header counts, for example showing extra idle sessions that are no longer actionable.
- Some rows may show `Unknown` as the terminal app when restored metadata no longer has a jump target.

## Suspected Cause

`AppModel.computeSessionBuckets()` builds the primary island list from every session where `session.isVisibleInIsland` is true.

For non-hook Claude sessions, `AgentSession.isVisibleInIsland` can remain true while `isProcessAlive` is true. If the Claude daemon / background process stays alive after the task is completed, stale completed sessions can stay in the primary list instead of aging out.

## Fix Direction

Keep the process-liveness state intact, but filter stale completed sessions out of the primary island list after `completedStaleThreshold`.

They should still be allowed to appear in recent / overflow history. The `.never` completed-stale threshold should continue to preserve completed sessions for users who explicitly want that behavior.

Octane0411:
## Summary

Fixes #576.

Completed sessions that have crossed the configured stale threshold now leave the primary island session list instead of staying counted as live just because their daemon or terminal process is still alive. They remain available through `recentSessions` / overflow history.

## Root Cause

`AppModel.sessionBuckets` cached the primary/overflow split even though the split depended on wall-clock time through completed-session staleness. A completed session could therefore stay in primary until some unrelated state mutation invalidated the cache.

## Changes

- Filter primary session buckets with the completed-stale threshold, so stale completed rows move to recent history.
- Add a time-aware session bucket cache with an expiry task so closed-island counts and agent slots refresh when a stale boundary is crossed without any state change.
- Use the same reference date for state grouping and bucket filtering.
- Add regression coverage for stale completed sessions leaving primary and preserving recent history.

## Validation

- `git diff --check`
- `swift build`

Targeted `swift test --filter 'AppModelSessionListTests|AgentsGridRightSlotTests'` could not run in this local CLT environment because test targets fail to compile with `no such module 'Testing'`, even with `--enable-swift-testing`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session lists now update more consistently as time passes, so completed items move between active and recent views at the expected threshold.
  * Stale completed sessions are handled more predictably when grouping and sorting islands.
  * Improved internal timing consistency makes session display behavior less likely to shift unexpectedly.

* **Tests**
  * Updated automated checks to use a fixed time reference for more reliable session-list behavior coverage.

### DIFF
--- a/Sources/OpenIslandApp/AgentSession+Presentation.swift
+++ b/Sources/OpenIslandApp/AgentSession+Presentation.swift
@@ -16,7 +16,7 @@ enum IslandSessionPresence: Equatable {
 
 extension AgentSession {
     private static let collapsedDetailAgeThreshold: TimeInterval = 20 * 60
-    private static let islandActivityThreshold: TimeInterval = 20 * 60
+    static let islandActivityThreshold: TimeInterval = 20 * 60
     static let staleCompletedDisplayThreshold: TimeInterval = 5 * 60
 
     /// Whether this session represents a subagent (worktree agent) that should
--- a/Sources/OpenIslandApp/AppModel.swift
+++ b/Sources/OpenIslandApp/AppModel.swift
@@ -33,6 +33,13 @@ final class AppModel {
     private static let liveSessionStalenessWindow: TimeInterval = 15 * 60
     private static let jumpOverlayDismissLeadTime: Duration = .milliseconds(20)
     private static let agentsGridObservedSequenceLimit = 512
+    private static let sessionBucketPriorityAgeBoundaries: [TimeInterval] = [
+        120,
+        900,
+        AgentSession.islandActivityThreshold,
+        3_600,
+        21_600,
+    ]
     static let hoverOpenDelay: TimeInterval = 0.15
 
     struct AcceptanceStep: Identifiable {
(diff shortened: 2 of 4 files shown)


Octane0411:
Thanks for the work here. We re-evaluated this behavior against the product semantics: a session should remain visible while its terminal/window is still alive. Aging every completed session out after a fixed elapsed time can therefore hide a valid, still-open session.

The stale-row problem in #576 is still worth solving, but removal should be driven by terminal/jump-target liveness and the existing timeout path rather than completion age alone. We have removed this implementation from the release branch and will not merge it in its current form, so I’m closing this PR while leaving #576 open for a liveness-aware solution.

Issue #507, opened 2026-05-24
Add setting to control or disable auto-collapse duration for OpenCode notification cards


emreinci987:
Hi, thanks for building Open Vibe Island.

I am using Open Vibe Island with OpenCode on macOS, installed via Homebrew. The OpenCode plugin seems to be installed correctly, and I can briefly see the agent output / terminal activity in the notch bar.

However, the notification/completion card disappears very quickly, often before I have enough time to read the text. This makes it hard to follow what the agent is doing, especially when OpenCode prints longer messages or when multiple sessions are running.

Current behavior:
- OpenCode is running in VS Code terminals.
- Open Island briefly shows the agent output in the notch/notch bar.
- After a short time, the card disappears automatically.
- I could not find a setting to increase this duration or keep the card visible.
- The “Auto collapse on mouse exit” behavior also does not seem to be configurable from the settings UI.

Expected / requested behavior:
It would be very helpful to have a setting such as:
- Notification card duration: 10s / 30s / 60s / Never
- Disable auto-collapse for OpenCode completion cards
- Keep latest OpenCode output visible until manually dismissed
- Pin completion/session card while the agent is active

My use case:
I usually run multiple OpenCode sessions from VS Code terminals. I want to use Open Vibe Island to quickly monitor what the agents are doing, but currently the text disappears too quickly to read comfortably.

Environment:
- macOS version: [your macOS version]
- Open Vibe Island installation method: Homebrew
- Open Vibe Island version: [version if available]
- OpenCode version: [opencode --version]
- Terminal: VS Code integrated terminal

Question:
Is there already a recommended way to keep OpenCode notification/completion cards visible for longer, or would you consider adding a setting for the auto-collapse duration?

Thanks!

PR #588, opened 2026-07-09
fix: keep Qoder sessions alive while a qoder process exists

imlx:
ProcessMonitoringCoordinator.sessionIDsWithAliveProcesses had explicit liveness strategies for every supported agent except Qoder. Because Qoder sessions are hook-managed (Claude Code fork, same hook format at ~/.qoder/settings.json), the default 'not in aliveSessionIDs' path incremented processNotSeenCount on every poll and flipped isSessionEnded=true after two polls. Result: the Qoder session vanished from the island within ~2-4 seconds, even while the qoder CLI was still running.

Additionally, ActiveAgentProcessDiscovery had no isQoderProcess matcher, so the qoder binary (installed by @qoder-ai/qodercli) was never discovered as an active agent process.

Adopt the same strategy used for Kimi CLI (commit 201f2b1): as long as any qoder process is observed by ActiveAgentProcessDiscovery, mark every tracked Qoder session as alive. Qoder's UUID session IDs are not recoverable from ps/lsof, and assuming one agent-per-process is good enough for v1.

Related to #507.

## Summary by CodeRabbit

* **New Features**
  * Added support for detecting Qoder CLI processes and capturing their session details.
  * Qoder sessions can now be recognized as active when only related processes are available.

* **Bug Fixes**
  * Improved process matching so Qoder is correctly identified and not misclassified as other supported tools.
  * Enhanced Qoder session liveness tracking to reduce false “inactive” states.

* **Tests**
  * Added coverage for discovering Qoder CLI processes and validating the resulting snapshot details.

### DIFF
--- a/Sources/OpenIslandApp/ActiveAgentProcessDiscovery.swift
+++ b/Sources/OpenIslandApp/ActiveAgentProcessDiscovery.swift
@@ -211,6 +211,23 @@ struct ActiveAgentProcessDiscovery {
                 ))
                 continue
             }
+
+            if isQoderProcess(command: process.command) {
+                let claimKey = "qoder:\(process.pid)"
+                guard claimedKeys.insert(claimKey).inserted else {
+                    continue
+                }
+
+                let lsofOutput = lsofOutput(pid: process.pid)
+                snapshots.append(ProcessSnapshot(
+                    tool: .qoder,
+                    sessionID: nil,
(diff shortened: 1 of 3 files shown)


imlx:
Closing to investigate root cause - the fix works for CLI process detection but Qoder IDE sessions still disappear when the model finishes responding. Need to understand what hooks Qoder actually sends (SessionEnd vs Stop) before re-submitting.

PR #604, opened 2026-07-21
fix: improve Codex session titles

simon-ami:
## Summary

- ignore Codex-injected `<recommended_plugins>` and pathless AGENTS context when selecting the initial user prompt
- use Codex's generated thread name for session headlines while keeping the workspace identity for grouping and jump-back
- reconcile generated and renamed titles from `session_index.jsonl` independently of large rollout transcripts, and consume live app-server rename notifications
- preserve live session state and cached titles while preventing overlapping rollout rediscovery passes

## Test Plan

- `swift test --filter CodexSessionTrackingTests`
- focused `SessionStateTests`, `AppModelSessionListTests`, and `AgentSessionPresentationTests` title regressions
- `swift build`
- manual verification with Codex.app thread generation and rename behavior
- `swift test` *(title regressions pass; the full run still reports pre-existing shared-preference failures in three appearance tests and one flaky AgentsGrid observation test)*

## Summary by CodeRabbit

- **New Features**
  - Codex session titles now preferentially reflect the latest thread name when available.
  - Thread-title updates can appear live during an active session, without requiring a full refresh.
  - Session discovery can populate Codex thread names from the session index data.

- **Bug Fixes**
  - Preserved Codex thread names and session context across metadata updates and rediscovery.
  - Spotlight headlines now use thread names for Codex while keeping workspace names for other tools.
  - Improved recognition of additional injected prompt content during session tracking.

### DIFF
--- a/Sources/OpenIslandApp/AgentSession+Presentation.swift
+++ b/Sources/OpenIslandApp/AgentSession+Presentation.swift
@@ -131,6 +131,16 @@ extension AgentSession {
         return trimmedTitle
     }
 
+    var spotlightSessionName: String {
+        if tool == .codex,
+           let threadName = codexMetadata?.threadName?.trimmedForSurface,
+           !threadName.isEmpty {
+            return threadName
+        }
+
+        return spotlightWorkspaceName
+    }
+
     var spotlightWorktreeBranch: String? {
         // This is a SwiftUI computed property read on every layout
         // pass. It MUST stay free of filesystem IO. Calling
@@ -168,7 +178,7 @@ extension AgentSession {
     }
 
     var spotlightHeadlineText: String {
-        var headline = spotlightWorkspaceName
+        var headline = spotlightSessionName
 
         if let branch = spotlightWorktreeBranch {
             headline += " (\(branch))"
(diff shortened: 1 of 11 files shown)


Octane0411:
Thanks — thread names from `session_index.jsonl` plus `threadNameUpdated` are a real improvement (built; merged onto current main 377 tests pass). Before merge:

1. Please rebase after #611 lands. The two conflict in `SessionDiscoveryCoordinator.swift`, `CodexSessionTracking.swift`, `AgentSession+Presentation.swift` and their tests. #611 carries the superset injected-marker classifier and the `## My request for Codex:` extraction, so this PR's `isInjectedPromptBlock` additions and the `discovered.initialUserPrompt ?? existing…` precedence flip (around line 344 — it would override hook-sourced prompts and contradicts #611's "replace only if the cached value is injected" rule) should be dropped; keep the thread-name work.
2. `SessionDiscoveryCoordinator.swift:442-461`: a 2-second infinite loop re-reads the whole append-only index for the app's lifetime whenever any Codex session exists. Given #631 was a perf fix, please use a `DispatchSource` / FSEvents watcher (see `CodexRolloutWatcher`) or at least an mtime check. Also `onStatusMessage?("Updated N Codex session title(s).")` at :537 lands in `lastActionMessage` on every rename — probably not intended.
3. Consider routing the rename through `SessionState.apply(_:)` via an event instead of the `onThreadNameUpdated` side channel + wholesale `state = SessionState(sessions:…)` (CLAUDE.md's single-source rule). Not blocking, since discovery already does this elsewhere.
4. `CodexAppServerCoordinator.swift:48`: the new `NSRunningApplication` guard in `ensureConnected()` is a behaviour change (no longer spawns the bundled app-server when Codex.app isn't running) — please mention it in the description.

Issue #607, opened 2026-07-21
[Bug / Enhancement] Does not select proper workspace/tab in cmux


pm0u:
When clicking a session that is in cmux, the window is foregrounded but the workspace that the session is in is not selected, nor the tab within a workspace.

OJ7:
I found this setting in cmux that gets this somewhat working:

<img width="740" height="166" alt="Image" src="https://github.com/user-attachments/assets/ade23f80-bdf5-4e11-881a-122ddd855758" />

It was set to `cmux processes only` for me. When I changed it, clicking on the session opens the correct workspace. If the correct tab isn't already active, clicking the session again selects the correct tab. Prior to this, I had the same behavior as @pm0u where only the cmux window would get activated (to whatever workspace/tab was last active).

The clue that lead me to this was this from the README in this repo: `cmux: Jump-back via Unix socket API`

PR #608, opened 2026-07-21
fix: cmux jump-back selects the target tab and workspace

pm-x:
`jumpToCmuxTerminal` sends `surface.focus` without `select:true`, so cmux moves focus to the surface but leaves whatever tab was already visible in the pane showing, and doesn't bring the surface's workspace forward. Jumping back lands you in the right window but the wrong tab/workspace.

Adding `"select":true` to the params fixes both. Tested against a live cmux socket: with plain `surface.focus` the pane's `selected_surface_id` doesn't change; with `select:true` it flips to the target surface *and* the active workspace switches to match - so for the single-window case the separate `workspace.select`/`window.focus` calls aren't needed. I didn't test a multi-window setup, so `window.focus` on the returned `window_id` might still be worth adding there - happy to follow up if a maintainer wants it.

**Test plan** (needs a real cmux, so CI can't cover it):
- Put two tabs in one cmux workspace plus a second workspace. Focus the *other* tab and switch to the *other* workspace.
- Trigger jump-back to a session in the first workspace → it now lands on the correct tab in the correct workspace, not just the right window.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal navigation so the selected terminal is now brought forward and displayed as the active tab in its pane.

### DIFF
--- a/Sources/OpenIslandApp/TerminalJumpService.swift
+++ b/Sources/OpenIslandApp/TerminalJumpService.swift
@@ -530,8 +530,10 @@ struct TerminalJumpService {
         }
         guard connectResult == 0 else { return false }
 
-        // Send JSON-RPC surface.focus request.
-        let request = #"{"jsonrpc":"2.0","method":"surface.focus","params":{"surface_id":"\#(surfaceID)"},"id":1}"# + "\n"
+        // Send JSON-RPC surface.focus request. `select:true` makes the surface
+        // the visible tab in its pane and brings its workspace forward — without
+        // it, cmux only moves focus and leaves whatever tab/workspace was showing.
+        let request = #"{"jsonrpc":"2.0","method":"surface.focus","params":{"surface_id":"\#(surfaceID)","select":true},"id":1}"# + "\n"
         let sent = request.withCString { ptr in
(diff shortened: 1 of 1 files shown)


pm-x:
Fixes #607

Octane0411:
Thanks for the quick fix. I couldn't confirm the mechanism, though: cmux's socket-API docs list `surface.focus` params as `surface_id` + optional `workspace_id`, and `ControlCommandCoordinator+Surface.swift::surfaceFocus` at cmux HEAD never reads a `select` key (no hits for it in the manaflow-ai/cmux history either). At HEAD `controlSurfaceFocus` already does `focusMainWindow` → `selectWorkspace` → `focusPanel`, so `select: true` looks like a silent no-op unless you're on a cmux build that parses it.

Could you share the cmux version where you saw the difference? If it's an older build, the documented pattern is `workspace.select {workspace_id}` followed by `surface.focus {surface_id}` — Open Island only captures `CMUX_SURFACE_ID` today, so the real fix for #607 is probably to also capture `CMUX_WORKSPACE_ID` in the hook payloads' runtime context and pass it along. Happy to merge a version that does that.

PR #611, opened 2026-07-22
fix: prevent injected Codex context from becoming session titles

qyy3369-hue:
## Summary

- ignore Codex-injected plugin, repository, environment, attachment, and browser-context blocks when deriving user prompt previews
- extract the actual request that follows `## My request for Codex:` from wrapped desktop-app messages
- replace stale cached prompt previews only when the cached value is known injected context
- avoid showing the root workspace placeholder (`/`) when a usable prompt title exists

## Problem

Codex desktop rollouts can store system-provided context alongside the user's request. Open Island treated the first available block as the session topic, so the island could display headings such as `<recommended_plugins>`, `# Files mentioned by the user`, or other wrapper metadata instead of the user's request.

### Visual reproduction

Before this fix, injected plugin metadata could become the visible session headline:

Open Island showing recommended_plugins as Codex session titles before the fix

Expected behavior: the headline should use the extracted user-authored request and omit injected wrapper metadata.

## Root cause

Prompt extraction skipped a small set of whole injected blocks, but it did not recognize newer Codex wrapper formats or extract the user-authored portion of a combined message. Persisted sessions could also retain an already-cached injected preview after rediscovery.

## Validation

The following targeted tests pass:

- `codexRolloutReducerTracksMessageResponsePromptsWithoutInjectedBlocks`
- `headlineOmitsRootWorkspacePlaceholderWhenPromptExists`
- `mergeDiscoveredCodexSessionsReplacesInjectedCachedInitialPromptOnly`

This PR is limited to prompt/title parsing and its regression coverage. It does not change Codex app-server startup, network configuration, or thread-name discovery.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session headline generation: root workspace placeholders no longer show when a prompt exists.
  * Refined Codex prompt extraction and injected-content filtering to surface the actual user request, including updated handling for multiple injected markers and message segments.
  * Enhanced discovery/merge behavior so injected cached initial prompts are correctly replaced (while valid cached prompts are preserved).
* **Tests**
  * Added headline coverage for root workspace behavior.
  * Added coverage for merging discovered Codex session prompts and updated injected-block prompt tracking fixtures.

### DIFF
--- a/Sources/OpenIslandApp/AgentSession+Presentation.swift
+++ b/Sources/OpenIslandApp/AgentSession+Presentation.swift
@@ -168,7 +168,8 @@ extension AgentSession {
     }
 
     var spotlightHeadlineText: String {
-        var headline = spotlightWorkspaceName
+        let workspaceName = spotlightWorkspaceName
+        var headline = workspaceName
 
         if let branch = spotlightWorktreeBranch {
             headline += " (\(branch))"
@@ -176,6 +177,10 @@ extension AgentSession {
 
         guard let prompt = spotlightHeadlinePromptText else {
             return headline
+        }
+
+        guard workspaceName != "/" else {
+            return prompt
         }
 
         return "\(headline) · \(prompt)"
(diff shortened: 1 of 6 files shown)


qyy3369-hue:
Hi @Octane0411, just checking whether this focused fix is still useful. I rechecked the current `main` after #631 / v1.1.7, and the additional injected wrappers plus `## My request for Codex:` extraction still appear to be unhandled. #611 remains mergeable, and CodeRabbit has no actionable comments. I am happy to rebase or adjust it if preferred, especially given the overlap with #613. Thanks!

Issue #602, opened 2026-07-16
Fix mess with recent changes to the app and codex


imkebe:
At the moment with the recent version of the app - 1.1.6 and recent Codex 26.707.91948 the app show subagents and under the hood calls to APi's making the Codex based usage a mess.

Despite running two chats (which i can see in the OVI) there are also "/" chats that are in app permission risk assesstment and similar under the hood calls. This make a flood of a agent notifications  in the app. Also background chats are shown but not avaible anymore in the Codex app making the focus to produce 

<img width="2478" height="189" alt="Image" src="https://github.com/user-attachments/assets/dbea631a-ce8c-482d-99c5-8d9bbd87ee18" />

"Unable to resume the task. Error resuming thread: Fatal error: Failed to initialize session: invalid thread-store request: thread 019f6b59-1432-7d81-88c4-4bd35bd5bccd already has a live local writer."

This might be in conflict with #584 and #579

goace:
### Additional reproducible evidence (still present in v1.1.7)

I reproduced the same hidden-session behavior and traced the affected rows back to Codex rollout metadata.

**Environment**

- Open Island: 1.1.7 (build 77)
- Codex session originator: `Codex Desktop`
- Codex CLI version recorded in the rollouts: `0.146.0-alpha.9.2`

**Observed**

Five idle rows appeared in Open Island even though none of them was a user-visible Codex task. In Open Island's persisted registry, all five were:

- `phase = completed`
- `attachmentState = stale`
- `terminalApp = Codex.app`

Each row mapped to a real file under `~/.codex/sessions/.../rollout-*.jsonl`. The relevant, sanitized `session_meta` field was identical for all five:

```json
{
  "originator": "Codex Desktop",
  "source": {
    "subagent": {
      "other": "guardian"
    }
  }
}
```

So these are internal Codex Desktop `guardian` sessions (for example, background permission/risk assessment), not conversations created or shown to the user. This also explains the root-like `/` rows and internal prompt text appearing as a title/summary.

**Likely cause**

[`parseSessionMeta(fromLine:)`](https://github.com/Octane0411/open-vibe-island/blob/main/Sources/OpenIslandCore/CodexSessionTracking.swift#L659-L679) currently reads only `id`, `cwd`, and `timestamp` from the payload. It does not inspect or retain `payload.source.subagent`, so rollout rediscovery treats internal guardian rollouts as normal Codex.app sessions.

[...]

Issue #531, opened 2026-06-05
Codex usage and session list display settings are incomplete


Ji-Sir:
## Problem
Usage display can flash 0% or miss weekly limits, and the session list has too many/few rows without flexible user controls.

## Expected
Codex usage should prefer app-server/rateLimits data and keep a sane cached fallback. Session list display should support custom active-time windows and fixed counts.

## Proposed fix
- Parse app-server account rate limits and JSONL token_count fallbacks.
- Ignore all-zero placeholder usage snapshots.
- Cache last known good usage.
- Add custom active-window minutes and fixed-count display settings.
- Keep closed/open session counts aligned and hide Codex subagents from the main list.

Apex-Studio-He:
I can reproduce the 0% Codex usage display on Open Island 1.1.6, and I found a concrete cause that may help narrow this issue down.

### Environment

- macOS 15.6.1
- Apple M2
- Open Island 1.1.6, installed through the official Homebrew cask
- Codex Desktop
- ChatGPT Pro

### Actual behavior

At the same time:

- Codex Desktop showed **Weekly usage remaining: 85%**
- Open Island's compact header showed **Codex 7d 0%**

The local rollout data confirms that the main Codex value is correct:

```text
limit_id: codex
primary.used_percent: 15.0
primary.window_minutes: 10080
secondary: null
```

So the expected remaining value is 85%.

However, another active Codex Desktop rollout was also writing a separate rate-limit bucket:

```text
limit_id: codex_bengalfox
primary.used_percent: 0.0
primary.window_minutes: 10080
secondary: null
```

At the time of the screenshot, the newest rate-limit event across the active rollout files was the `codex_bengalfox` event, so Open Island displayed 0%. The main `limit_id: codex` bucket still contained 15% used / 85% remaining.

### Likely cause

`CodexUsageLoader.load()` selects the most recently modified rollout file containing a rate-limit snapshot, but it does not distinguish snapshots by `limit_id`. With multiple active Codex Desktop sessions writing different buckets, the displayed value can therefore flip depending on which rollout file was updated most recently.

This looks slightly different from a generic all-zero placeholder: the zero value belongs to a real but separate `codex_bengalfox` limit bucket.

### Expected behavior

For the generic “Codex” usage indicator, Open Island should preferably:

1. select `limit_id == "codex"`, or
2. retain/group separate limit buckets explicitly instead of allowing one to replace another, and
[...]

PR #613, opened 2026-07-23
Fix Codex session titles, usage display, and metadata

### LINKED ISSUE #610: Codex.app session rows show injected metadata instead of the thread name
## Problem

Codex.app sessions can show injected request metadata such as `<recommended_plugins>...` in the expanded session-row headline, even though the Codex task has a user-visible thread name.

## Reproduction

1. Start a named task in Codex.app where the first rollout `user_message` begins with an injected metadata block.
2. Open Open Island and expand the Codex session row.
3. The headline is derived from the initial prompt and displays something like `/ · <recommended_plugins>...` instead of the task name shown by Codex.app.

## Expected

For Codex.app sessions, prefer the real Codex thread name. Keep the prompt-based fallback for CLI sessions or unnamed threads.

## Technical notes

- `CodexThread.name` is already available from the app-server and is used when emitting the initial `SessionStarted` event.
- `thread/name/updated` notifications are parsed, but `CodexAppServerCoordinator` currently ignores them.
- `AgentSession+Presentation.spotlightHeadlineText` uses `initialPromptText ?? latestPromptText` and does not prefer the session title for Codex.app.
- `CodexRolloutReducer.applyUserMessage` records the first `user_message` verbatim, including leading injected metadata.

A narrow fix could add a title-update event for existing sessions, apply `thread/name/updated`, sync names from loaded threads, and prefer a non-generic Codex.app session title in row presentation.

Apex-Studio-He:
## Summary

- Display Codex quotas as remaining percentages and select the most constrained active window by usage.
- Synchronize authoritative Codex Desktop task titles, model, reasoning level, service tier, and exact thread identifiers.
- Replace duplicate app-source metadata with an adaptive single-line model/reasoning/service badge.
- Add a per-display setting to hide idle sessions, including consistent collapsed-pill counts.
- Make rollout discovery, title resolution, usage parsing, and elapsed timers incremental and cache-aware.
- Resume Codex Desktop threads after long idle periods without losing them during process reconciliation.

## Latest update

- Use Codex `timeUsedSeconds` snapshots as the authoritative goal duration instead of counting all wall-clock time since goal creation.
- Rebase goal, explicit plan-mode, and current-turn timers together when a newer authoritative counter arrives, excluding inactive gaps.
- Persist accumulated active duration separately from the currently running segment while retaining timestamp fallbacks for older rollouts.
- Migrate already-cached active sessions with a one-time bounded rollout backfill, then return to the normal small incremental polling window.
- Hide internal Codex child-agent threads from the top-level session list, leaving only the real parent task visible.
- Show only running sessions and genuinely recent completions when idle sessions are hidden.
- Preserve renamed conversation titles and completion recency while synchronizing project-backed tasks.
- Stop rollout observations from replaying the same transcript tail when cached prompt or timer metadata changes.
- Avoid full-row animation redraws while elapsed timers update.
- Drive standalone mouse-wheel scrolling from one display-synchronized 60 Hz loop, while leaving native trackpad momentum untouched.
- Keep Simplified Chinese, Traditional Chinese, and English labels consistent.

## Root cause

Codex app-server events, SQLite fallbacks, and rollout files were treated as independent sources. This allowed fallback names to overwrite authoritative task titles, dropped metadata for project-backed sessions, and repeatedly reparsed large rollout files. Internal child-agent rollouts use the same session format as user-created tasks, but their `source.subagent` marker was ignored, so every child was imported as a separate top-level task. Metadata-only synchronization could also refresh a completed session's activity timestamp, causing old history to reappear as a recent completion.

Elapsed timers had a related source-of-truth problem: the UI subtracted a saved start timestamp from the current time and ignored the authoritative goal counter emitted by Codex. Pauses, sleep, and other inactive gaps were therefore included in goal and task durations.

## User impact

Codex sessions now show their real task names and useful runtime metadata, deep-link to the exact conversation when an identifier is available, report remaining quota consistently, and can hide idle rows without corrupting summary counts. A task that runs several internal child agents remains one visible parent task instead of expanding into duplicate-looking rows. Completed cards keep the real task title, while old completions stay out of the visible recent list.

Goal, plan, and thinking timers now follow Codex's accumulated active duration and continue from the latest official update rather than displaying inflated wall-clock totals. Existing cached sessions migrate automatically without requiring users to clear their conversation history.

## Validation

- 38 `CodexSessionTrackingTests` passed, including authoritative-counter rebasing, cache persistence, and legacy timer backfill.
- 21 `AgentSessionPresentationTests` passed, including accumulated goal/plan/thinking timer presentation and timestamp compatibility fallback.
- Built, installed, and launched the development app, then verified a live long-running session migrated to the official goal and current-turn baselines.
- Previous full regression run passed 366 tests with the known grid-slot test excluded.
- Profiled the running development app during repeated scrolling: the main thread remained idle for about 99.3% of samples, and idle app CPU stayed near 0%.

Closes #610
Related to #531
Related to #602


imkebe:
Fixes #602 once done

Apex-Studio-He:
@imkebe Thanks for confirming! I’m glad these changes also address #602. I’ve included fixes for internal subagent filtering, stale session handling, and duplicate notifications. If you find the improvements useful, I’d really appreciate a star on [my fork](https://github.com/Apex-Studio-He/open-vibe-island). ⭐

Apex-Studio-He:
## Update: restore active goals after cache loss

- Detect current goal-continuation events even when legacy/local cache entries no longer contain goal start or timer fields.
- Trigger the existing bounded history backfill only for affected sessions, then restore the latest authoritative `createdAt`, accumulated duration, and running state.
- Keep internal goal context out of the visible user prompt while preserving normal cached task text.

### Root cause

The recovery guard only requested a deeper rollout scan when a cached goal start already existed. If both cached goal fields were missing, the current continuation signal was ignored and the session displayed only its thinking timer.

### Validation

- 39/39 `CodexSessionTrackingTests` passed.
- Reproduced the missing-goal state with a legacy cache fixture and verified the restored timer behavior.
- Installed the development build and confirmed the affected live PCB session recovered its goal start and authoritative accumulated duration.

VihaanMotwani:
Tested the latest head (`3938969`) locally and ran into one startup edge case.

If one Codex task is already live and another running task is discovered from its rollout when Open Island starts, the discovered task gets removed after the second reconciliation pass.

I added a focused regression test and tested both versions with only the mapping changed. With:

```swift
payload.discoveredCodexRecords.map(\.session)
```

the test ends up with one task (`liveSessionCount == 1`). Using:

```swift
payload.discoveredCodexRecords.map(rediscoveredSession(from:))
```

keeps both running tasks and the test passes (`liveSessionCount == 2`).

It looks like the startup path is missing the Codex.app/process identity setup. I re-ran this after the latest active-goal continuation commit and the behavior is unchanged, so that change seems separate.

Happy to open a small PR with the test if useful.

Apex-Studio-He:
@VihaanMotwani Thank you for taking the time to test the latest head and isolate this so carefully. I reproduced the same second-reconciliation behavior locally, and your `rediscoveredSession(from:)` mapping looks like the right fix.

Yes, please open the small PR with the focused regression test against `Apex-Studio-He/open-vibe-island:fix/codex-usage-and-titles`. I’d be happy to review and incorporate it. Really appreciate your help with this.

PR #615, opened 2026-07-23
feat: add CLI commands for the Claude usage bridge

namearth5005:
## Problem

The Claude **usage bridge** (the managed status line that tees `rate_limits` into `/tmp/open-island-rl.json`, feeding the notch usage panel) could only be installed from the app's **Settings window**. That leaves two gaps:

- No **headless / scriptable** install — you can't set it up from a terminal, a dotfiles script, or a provisioning step, unlike the hook installers (`installClaude`, `installKimi`, …).
- Users with a **custom status line** (e.g. a Claude Desktop workflow that also runs terminal `claude`) had no obvious path: the docs said the bridge "refuses to overwrite an existing custom status line," even though the manager already supports a non-destructive **wrapper mode**.

This matters for Desktop-app users specifically: Desktop's headless `stream-json` mode never renders a status line, but rate limits are account-wide, so one terminal `claude` render through the bridge seeds the cache and the Desktop pill then shows Claude usage. Getting the bridge installed (in wrapper mode, preserving the user's status line) is the enabler.

## Change

- Add `installClaudeUsage` / `uninstallClaudeUsage` / `statusClaudeUsage` to `OpenIslandSetup`, mirroring the existing hook-installer commands.
- Extract the "install, auto-wrapping an existing custom `statusLine`" decision into `ClaudeStatusLineInstallationManager.installPreservingExisting()`, so the app UI and the CLI share **one** code path. The app's `installClaudeUsageBridge()` now calls it instead of duplicating the `try install()` / `catch → installAsWrapper()` fallback.
- `installClaudeUsage` reports whether it landed in managed or wrapper mode; `uninstallClaudeUsage` restores the original status line; `statusClaudeUsage` prints managed/wrapper/repair state.

No behavior change for existing users — the app's Settings action produces the identical result (it now routes through the shared method).

## Testing

- Two new `OpenIslandCore` tests for `installPreservingExisting`: wraps when a custom `statusLine` is present (and the original command still runs via the delegate); uses the plain managed script when none exists. Full `ClaudeUsageTests` suite green (10/10).
- Verified end-to-end on a real machine with a custom status line: `installClaudeUsage` installed in wrapper mode (original `statusLine` preserved under `_openIslandOriginalStatusLine`), and piping a status-line payload through the installed script wrote a correctly-shaped `/tmp/open-island-rl.json` while still running the user's original command.
- `scripts/check-docs.sh` passes; README updated (CLI commands + corrected wrapper behavior).

> Note: `AppModelSessionListTests` has 4 pre-existing failures on a clean `main` (unrelated to this change, verified separately).

---

### 中文摘要

Claude **用量桥接**（写入 `/tmp/open-island-rl.json` 的托管 status line，为 notch 用量面板供数）此前只能从应用的设置窗口安装，无法在终端中无头安装或脚本化；且对已有自定义 status line 的用户（如同时使用终端 `claude` 的 Claude 桌面工作流）不友好。

本 PR 为 `OpenIslandSetup` 新增 `installClaudeUsage` / `uninstallClaudeUsage` / `statusClaudeUsage` 命令，与现有 hook 安装命令对齐；并将"安装并在存在自定义 status line 时自动进入 wrapper 模式"的判定抽取为 `ClaudeStatusLineInstallationManager.installPreservingExisting()`，供应用 UI 与 CLI 共用同一路径。由于额度是账户级的，终端 `claude` 的一次渲染即可通过桥接为缓存供数，桌面会话的用量随之显示。新增两个测试并更新 README。

## Summary by CodeRabbit

* **New Features**
  * Claude usage setup now installs in wrapper mode when a custom status line already exists, preserving the user’s current `statusLine.command`.
  * Added `installClaudeUsage`, `uninstallClaudeUsage`, and `statusClaudeUsage` subcommands.
  * Status reporting now includes wrapper/repair state plus the current status line (when present) and the rate-limit cache location.
* **Documentation**
  * Updated Claude usage setup documentation and CLI instructions to reflect the new opt-in wrapper behavior and reporting details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,7 +343,13 @@ swift run OpenIslandSetup uninstall
 
 #### Connect Claude Code
 
-Claude usage setup is available from the app's Settings window and remains opt-in. The bridge writes a managed `statusLine.command` to `~/.open-island/bin/open-island-statusline`, caches `rate_limits` into `/tmp/open-island-rl.json`, and refuses to overwrite an existing custom status line automatically.
+Claude usage setup is available from the app's Settings window and remains opt-in. The bridge writes a managed `statusLine.command` to `~/.open-island/bin/open-island-statusline` and caches `rate_limits` into `/tmp/open-island-rl.json`. When you already have a custom status line, it installs in **wrapper mode** — your original command keeps running unchanged while the bridge tees `rate_limits` to the cache. Because rate limits are account-wide, one terminal `claude` render seeds the cache and the panel then reflects total usage (including Desktop sessions). Manage installation from the Settings window, or via CLI:
+
+```zsh
+swift run OpenIslandSetup installClaudeUsage    # install; auto-wraps an existing custom status line
+swift run OpenIslandSetup statusClaudeUsage     # report managed / wrapper / repair state
+swift run OpenIslandSetup uninstallClaudeUsage  # remove and restore the original status line
+```
 
 ### Repository Map
 
--- a/Sources/OpenIslandApp/HookInstallationCoordinator.swift
+++ b/Sources/OpenIslandApp/HookInstallationCoordinator.swift
@@ -1051,13 +1051,10 @@ final class HookInstallationCoordinator {
 
     func installClaudeUsageBridge() {
         updateClaudeUsageBridge(userMessage: "Installing Claude usage bridge.", intent: .installed) { manager in
-            do {
-                return try manager.install()
-            } catch ClaudeStatusLineInstallationError.existingStatusLineConflict {
-                // User already has a custom statusLine (e.g. claude-hud). Install as a
-                // wrapper so their script keeps running and we still get rate_limits.
-                return try manager.installAsWrapper()
-            }
+            // Managed install, auto-wrapping any existing custom statusLine
+            // (e.g. claude-hud) so the user's script keeps running and we still
+            // capture rate_limits. Shared with the OpenIslandSetup CLI.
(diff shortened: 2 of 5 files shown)


Octane0411:
Thanks — the CLI commands, help text and README changes look good and consistent with the rest of `OpenIslandSetup`. Please rebase once #676 lands: it fixes the fork bomb from #671 and rewrites `install()` / `installAsWrapper()` / `uninstall()` / `status()` in the same file.

- `installPreservingExisting()`'s `if status().managedStatusLineIsWrapper { return try installAsWrapper() }` shortcut routes a poisoned wrapper into `installAsWrapper()`, which under #676 throws `wrapperWouldRecurse` instead of repairing (and without #676 re-wraps the poisoned original). After #676, `install()` itself rebuilds a healthy wrapper and repairs a poisoned one, so the helper can collapse to `do { try install() } catch existingStatusLineConflict { try installAsWrapper() }`, and this PR's `installAsWrapper()` changes should be dropped in favour of #676's.
- `statusClaudeUsage` should print the new `managedStatusLineIsPoisoned`; `uninstallClaudeUsage` prints "Restored your original statusLine command." on `wasWrapper`, but after #676 uninstalling a poisoned wrapper clears the status line — gate the message on post-uninstall `status.hasStatusLine`.
- Optional: `--claude-dir` still writes the real `~/.open-island/bin` script; a `--script-dir` flag would give parity with `--hooks-binary`.

PR #617, opened 2026-07-24
refactor: drive usage pills from a provider registry

### LINKED ISSUE #501: CC 用量 & 刷新时间 显示不完整
Image
当前使用量 以及 刷新时间。

Image
展示的用量不对，以及刷新时间未展示

1AdityaX:
Closes #501.

Overlaps with #613. That PR restructures the same UsageProviderPresentation and UsageWindowPresentation types in IslandPanelView.swift to show Codex usage as remaining percentage instead of used percentage. This PR takes a different approach (a UsageProvider registry instead of string IDs) and also fixes stale/expired windows still being displayed, which #613 does not address. Whichever merges first, the other needs a rebase.

## Problem

Usage was written as two hardcoded pipelines. Provider identity lived in raw "claude" / "codex" strings that had to agree across the selection helper, the pill IDs, and a shortTitle switch. The island rebuilt each snapshot shape by hand, summary text was duplicated per provider, and the Codex toggle only persisted, so switching it on mid-session left the pill blank until the next launch.

Usage caches are also only rewritten while a harness runs, so a cache that outlives its window kept reporting the previous window's number. A 12h-old Claude cache whose 5h window had reset 7h earlier still headlined the pill as "5h 10%", reading like the weekly figure in the wrong slot, since the peak picker promotes the stale value over the live 7d window.

## Changes

- UsageProvider is the single registry now: labels, which harness draws down the quota, the Settings opt-in key, poll interval.
- UsageSnapshotSummarizing (Core) normalizes Claude's five_hour/seven_day and Codex's primary/secondary into one UsageWindowSummary shape.
- Windows past their resets_at are dropped from both the pill and the Settings summary. A provider disappears once every window has rolled over instead of showing dead numbers.
- Toggling a provider in Settings starts or stops its poll loop immediately instead of only persisting.

## Testing

swift build, swift test. Pre-existing failures in AppModelSessionListTests and AgentsGridRightSlotTests reproduce identically against a clean main with none of this branch's changes applied. Unrelated flakiness, not introduced here.

## Summary by CodeRabbit

* **New Features**
  * Added support for usage monitoring across multiple providers with individual opt-in settings.
  * Added a provider selector to switch between available usage summaries.
  * Usage displays now show active limits, reset times, and stale-data indicators.
  * Provider selection can follow the active tool or a manual override.

* **Bug Fixes**
  * Usage windows now disappear after their reset time.
  * Improved session detail and notification badge presentation.

* **Tests**
  * Added coverage for provider selection, opt-in behavior, usage expiry, and summary formatting.

### DIFF
--- a/Sources/OpenIslandApp/AppModel.swift
+++ b/Sources/OpenIslandApp/AppModel.swift
@@ -20,7 +20,6 @@ final class AppModel {
     private static let hapticFeedbackEnabledDefaultsKey = "app.hapticFeedbackEnabled"
     private static let islandRightSlotDefaultsKey = "appearance.island.v6.rightSlot"
     private static let islandCenterLabelDefaultsKey = "appearance.island.v6.centerLabel"
-    private static let showCodexUsageDefaultsKey = "app.showCodexUsage"
     private static let completionReplyEnabledDefaultsKey = "feature.completionReply.enabled"
     private static let suppressFrontmostNotificationsDefaultsKey = "app.suppressFrontmostNotifications"
     private static let legacyIslandSessionStateIndicatorDefaultsKey = "appearance.island.v8.stateIndicator"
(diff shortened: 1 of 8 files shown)


1AdityaX:
image



 Screenshot of the dropdown to see usage across multiple AI coding assistants

PR #636, opened 2026-08-04
feat: option to show remaining usage instead of used

Kevindic0214:
## Problem

Usage chips and their tooltips report consumption: a fresh 7-day window reads `7d 16%`. But a quota is watched for its **headroom**, not its consumption — the question at a glance is "how much do I have left before this window caps me", and today that means subtracting from 100 in your head every time you look at the notch.

## Change

Add an app-level toggle, **Show remaining quota instead of used**, in the Usage Bridge section of General settings — right beside the existing *Show Codex usage* toggle, which is the same kind of app-wide usage preference.

Off by default. Nothing changes for anyone who does not turn it on.

Two details worth calling out:

**Colour stays anchored to consumption.** `usageColor(for:)` keeps receiving `usedPercentage`, not the rendered number. Inverting the thresholds alongside the figure would make red mean "plenty left" and green mean "nearly capped" — exactly backwards. So a window at 95% used still shows red while displaying `5%`.

**The tooltip disambiguates.** The chip has no room to state which direction it counts, so `usageHelpText` does it: `5h 84% left` in remaining mode against a bare `5h 16%` in the default. New format key `island.usageWindowRemaining`.

`roundedRemainingPercentage` clamps to `0...100`: providers can report past 100% once a window is exhausted, which would otherwise render as a negative remainder.

This is provider-agnostic — Codex windows flip along with Claude's.

## Screenshots

Captured via `scripts/smoke-dev-app.sh` (`sessionList` scenario) against live `rate_limits` data — the same window, both directions.

**Default — used**

used

**Toggle on — remaining**

remaining

Note the colour stays green in both: 16% used is plenty of headroom either way you print it.

## Verification

- `swift build` — passes.
- `scripts/lint-strings.sh` — passes (all three locales carry both new keys).
- `scripts/check-docs.sh` — passes.
- Manual — harness smoke in both directions, screenshots above.
- `swift test` — **could not run locally.** This machine has only the Command Line Tools toolchain, which ships no `swift-testing` module; every test file fails with `no such module 'Testing'`. The identical failure reproduces on a pristine `main`, so it is an environment gap on my side rather than a regression. CI covers the actual test run.

## Notes

- Independent of #635 (detailed usage display) — that one changes *which windows* a chip lists, this one changes *which direction* a window counts. Both touch `compactUsageChip`, so whichever lands second will want a trivial rebase, but neither depends on the other.

---

### 中文摘要

用量膠囊顯示的是「已用」，所以 7 天視窗會讀成 `7d 16%`。但看額度時真正在意的是「還剩多少」，現在每次瞄一眼凹槽都得自己心算 100 減掉。

本 PR 在 General 設定的 Usage Bridge 區塊加一個 app 層級開關「顯示剩餘額度而非已用額度」，就放在既有的 Codex 用量開關旁邊。預設關閉，不開就完全不影響。

兩個刻意的取捨：顏色仍然由 `usedPercentage` 決定，不跟著數字一起翻 —— 否則紅色會變成「還很多」，語意整個顛倒。另外膠囊本身沒有空間標示方向，所以由 tooltip 講清楚：剩餘模式顯示 `5h 剩餘 84%`，預設模式維持 `5h 16%`。

## Summary by CodeRabbit

* **New Features**
  * Added a Settings toggle to display remaining Codex usage quota.
  * The island panel now shows remaining usage percentages when enabled.
  * Usage details adapt to available space by simplifying content as needed.
  * Session badges remain on a single line for improved readability.
  * Added English, Simplified Chinese, and Traditional Chinese translations.

### DIFF
--- a/Sources/OpenIslandApp/AppModel.swift
+++ b/Sources/OpenIslandApp/AppModel.swift
@@ -24,6 +24,7 @@ final class AppModel {
     private static let islandRightSlotDefaultsKey = "appearance.island.v6.rightSlot"
     private static let islandCenterLabelDefaultsKey = "appearance.island.v6.centerLabel"
     private static let showCodexUsageDefaultsKey = "app.showCodexUsage"
+    private static let usageShowsRemainingDefaultsKey = "app.usageShowsRemaining"
     private static let completionReplyEnabledDefaultsKey = "feature.completionReply.enabled"
     private static let suppressFrontmostNotificationsDefaultsKey = "app.suppressFrontmostNotifications"
     private static let legacyIslandSessionStateIndicatorDefaultsKey = "appearance.island.v8.stateIndicator"
@@ -261,6 +262,13 @@ final class AppModel {
         didSet {
             guard hasFinishedInit, showCodexUsage != oldValue else { return }
             UserDefaults.standard.set(showCodexUsage, forKey: Self.showCodexUsageDefaultsKey)
+        }
+    }
+    /// Renders usage windows as headroom left rather than quota consumed.
+    var usageShowsRemaining: Bool = false {
+        didSet {
+            guard hasFinishedInit, usageShowsRemaining != oldValue else { return }
+            UserDefaults.standard.set(usageShowsRemaining, forKey: Self.usageShowsRemainingDefaultsKey)
         }
     }
     var completionReplyEnabled: Bool = false {
@@ -621,6 +629,7 @@ final class AppModel {
                 atPath: CodexRolloutDiscovery.defaultRootURL.path
             )
         }
+        usageShowsRemaining = UserDefaults.standard.bool(forKey: Self.usageShowsRemainingDefaultsKey)
         completionReplyEnabled = UserDefaults.standard.bool(forKey: Self.completionReplyEnabledDefaultsKey)
         launchAtLoginEnabled = LaunchAtLoginService.shared.isEnabled
         appearanceSettingsProfile = IslandAppearanceDisplayProfile(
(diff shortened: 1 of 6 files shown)


Kevindic0214:
@coderabbitai full review

The previous run reported `Git: Failed to clone repository`, so the review was incomplete. Requesting a fresh pass.

Also addressed the English label in 91aa7aa — `settings.general.usageShowsRemaining` now reads "Show remaining quota instead of used quota", matching the zh-Hans/zh-Hant strings.

Octane0411:
Thanks — this is orthogonal to the other usage-chip PRs (which window is shown vs. which direction it counts) and worth having, but it conflicts textually with #675 (both rewrite `compactUsageChip`). Plan: #675 lands first; please rebase this on top — drop `peakPercentage(showingRemaining:)` and apply `roundedRemainingPercentage` per window inside the window `ForEach` (the tooltip change merges cleanly).

Two small things for the rebase:
- `Localizable.strings:202` uses `%d` with a Swift `Int`; repo convention is `%lld` (see `island.showAll`, `subagents.title`).
- The chip renders a bare number in both directions and only the tooltip says "left", so `7d 84%` in green is easy to misread — a short suffix when space allows would help.

Kevindic0214:
Rebased onto main (now includes #675). Applied the plan:

- Dropped `peakPercentage(showingRemaining:)`; `roundedRemainingPercentage` is now applied per window inside the `windows` `ForEach` in `compactUsageChip`, colour still keyed off `window.usedPercentage`.
- `island.usageWindowRemaining` now uses `%lld` in all three locales, matching `island.showAll` / `subagents.title`.
- Added a short "left" suffix to the chip's percentage text when the layout has room (`!layout.usesShortTitle`), so `7d 84%` doesn't read as consumption in the roomier tiers — the compact/short-title tier stays a bare number, disambiguated by the tooltip as before.

`swift build`, `scripts/lint-strings.sh`, and `scripts/check-docs.sh` all pass. `swift test` still can't run locally (no `swift-testing` module on this toolchain, same as noted originally).

Kevindic0214:
Fixed in 9e5222a: `UsageChipLayout` now conforms to `Sendable, Codable`, matching the repo convention. `swift build` passes.

PR #653, opened 2026-08-12
fix: detect our own hooks by ownership instead of "file would change"

css521:
## Summary

- split "a hook we should replace" from "a hook that is ours" — the same predicate was answering both, so a foreign hook read as a successful install
- derive `managedHooksPresent` from actual ownership in both the Claude and Codex installers, instead of from "the removal pass would change the file"
- add `HookHealthReport.Issue.hooksMissing` (error severity, auto-repairable) so a config another tool rewrote is visible and self-healing rather than silent
- gate that issue on the caller's persisted intent, so it never resurrects hooks a user deliberately removed
- regression coverage for both the installer and the health check

## The failure

Open Island can stop receiving hook events while reporting itself as installed and healthy.

Reproduced on a machine that also had the closed-source Vibe Island app installed. `~/.claude/settings.json` held **zero** `OpenIslandHooks` entries — a leftover `vibe-island-bridge` hook occupied the slots — and yet:

- Settings said *Claude hooks installed*
- the health panel showed only the informational `Other hooks coexist: …` notice, no error
- the startup repair never ran, even though the persisted intent was `installed`

Net effect: the island showed *"no open terminal sessions"* with a Claude Code agent running in the next terminal tab. Feeding one event straight into the bridge by hand produced a correct, fully-attached session row, which isolated the break to hook invocation rather than the bridge, `AppModel`, terminal resolution, or persistence.

## Root cause

One predicate answering two questions.

`ClaudeHookInstaller.isLegacyOpenIslandHookCommand()` matches `vibe-island-bridge`. That is right for *"drop this entry when we write our own hooks"* — two islands reacting to every event is not a supported state — but wrong for *"are our hooks installed?"*.

`status()` derived `managedHooksPresent` from that removal pass, so a foreign hook counted as ours. That false positive then propagated:

| | |
|---|---|
| `claudeHooksInstalled` | `true` → Settings shows the hooks as present |
| `shouldAutoInstall(.claudeCode)` | `.installed && !present` → `false`, so the repair written for exactly this case never fires |
| `HookHealthCheck.checkClaude` | no issue type for "none of our hooks are here", so nothing to report and nothing to repair |

`CodexHookInstallationManager.status()` had the same shape through `mutation.changed`, which is additionally true when the file merely re-serializes differently — we write sorted, pretty-printed JSON, so any `hooks.json` not already in that exact form read as installed.

## Approach

`isManagedHook()` stays broad; installing still claims the slots as before. A new `isOwnManagedHook()` / `ownHooksPresent` answers the ownership question, matching only the command we recorded at install time or our hook CLI by name — `OpenIslandHooks` and the pre-rename `VibeIslandHooks` — so an app bundle that moved since install still reads as ours.

`hooksMissing` is reported only when the caller passes `expectsInstalledHooks`. `repairHooksIfNeeded()` does not consult the intent store, so reporting it unconditionally would reinstall hooks a user turned off on purpose and regress #324.

## Behaviour change worth flagging

A user who only ever had the closed-source app's hooks is now recorded as `.untouched` rather than `.installed` by `migrateIntentStoreIfNeeded()`. They get first-run onboarding instead of a silent takeover, which seems like the better default, but it is a visible change for that group.

## Not covered here

`KimiHookInstaller`, `GeminiHookInstaller` and `CursorHookInstaller` carry the same broad-predicate shape. They are left alone because `HookHealthCheck` does not cover those agents, so there is no reporting path to fix alongside them, and I have no reproduction for them. Worth a follow-up.

## Verification

Passing locally:

- `scripts/lint-strings.sh` — the `lint` step of the CI harness
- `scripts/check-docs.sh` — the `docs` step
- `OpenIslandCore` and `OpenIslandCoreTests` **compile clean**, including both new test files. Six of the seven changed files live in `OpenIslandCore`, which has no external dependencies, so I built it through a throwaway `Package.swift` containing only that target and its tests.

Not run locally, both for environment reasons rather than anything about the change:

- **test execution** — swift-testing needs `Testing.framework`, which ships with Xcode, and this machine has only CommandLineTools. The bundle builds, then `dlopen` fails on the framework.
- **`swift build` of `OpenIslandApp`** — a cold worktree has to re-clone Sparkle and swift-markdown-ui, and this network moved about 2 MB in ten minutes. `HookInstallationCoordinator.swift` is the one changed file this leaves unbuilt.

Please treat the harness run as the gate. The run is currently sitting at `action_required`, as outside-contributor runs on this repo do, so it needs a maintainer to approve it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hook health checks to identify missing Open Island hooks when installation is expected.
  * Prevented legacy bridge hooks from being incorrectly reported as Open Island hooks.
  * Improved malformed configuration handling and avoided duplicate diagnostics.
  * Hook repair and verification now more accurately reflect installation status.

* **Tests**
  * Added coverage for missing, installed, malformed, and intentionally absent hook configurations.
  * Added regression coverage for legacy bridge hook cleanup.

### DIFF
--- a/Sources/OpenIslandApp/HookInstallationCoordinator.swift
+++ b/Sources/OpenIslandApp/HookInstallationCoordinator.swift
@@ -472,9 +472,17 @@ final class HookInstallationCoordinator {
             guard let self else { return }
 
             let binaryURL = self.hooksBinaryURL
+            let expectsClaude = self.expectsInstalledHooks(.claudeCode)
+            let expectsCodex = self.expectsInstalledHooks(.codex)
             let (claudeReport, codexReport, openCodeReport) = await Task.detached(priority: .utility) {
-                let claude = HookHealthCheck.checkClaude(hooksBinaryURL: binaryURL)
-                let codex = HookHealthCheck.checkCodex(hooksBinaryURL: binaryURL)
+                let claude = HookHealthCheck.checkClaude(
+                    hooksBinaryURL: binaryURL,
+                    expectsInstalledHooks: expectsClaude
+                )
+                let codex = HookHealthCheck.checkCodex(
+                    hooksBinaryURL: binaryURL,
+                    expectsInstalledHooks: expectsCodex
+                )
                 let openCode = HookHealthCheck.checkOpenCode()
                 return (claude, codex, openCode)
             }.value
@@ -500,9 +508,17 @@ final class HookInstallationCoordinator {
 
         // Re-run health checks first
         let binaryURL = hooksBinaryURL
+        let expectsClaude = expectsInstalledHooks(.claudeCode)
+        let expectsCodex = expectsInstalledHooks(.codex)
         let (claudeReport, codexReport, openCodeReport) = await Task.detached(priority: .utility) {
-            let claude = HookHealthCheck.checkClaude(hooksBinaryURL: binaryURL)
-            let codex = HookHealthCheck.checkCodex(hooksBinaryURL: binaryURL)
+            let claude = HookHealthCheck.checkClaude(
+                hooksBinaryURL: binaryURL,
+                expectsInstalledHooks: expectsClaude
+            )
+            let codex = HookHealthCheck.checkCodex(
+                hooksBinaryURL: binaryURL,
+                expectsInstalledHooks: expectsCodex
+            )
             let openCode = HookHealthCheck.checkOpenCode()
             return (claude, codex, openCode)
         }.value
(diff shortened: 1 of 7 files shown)


Octane0411:
Thanks — the ownership-based approach is the right fix, and the Claude side looks good (built + 374 tests pass on top of current main; legacy `VibeIslandHooks` entries still count as ours). One gap before merge:

- `CodexHookInstaller.swift:157` still derives `ownHooksPresent` from `containsManagedHook`, whose `isManagedHook` accepts `statusMessage == "Managed by Vibe Island"` and has no binary-name fallback. So (a) a foreign entry carrying the legacy status message still reads as installed, and (b) a legacy `…/VibeIslandHooks` entry without a manifest reads as *not* installed while `HookHealthCheck.hasOpenIslandHooks` says it is — two sources of truth disagree. Please add `containsOwnManagedHook` / `isOwnManagedHook` in the Codex installer mirroring the Claude one (exact `managedCommand` OR `openislandhooks` / `vibeislandhooks` substring, ignoring the status message), plus a regression test with a status-message-only entry. This is also CodeRabbit's first point; the follow-up commit was docs-only.

Nits (non-blocking):
- `HookInstallationCoordinator.swift:820-829`: the tail of `shouldAutoInstall`'s doc comment ended up above `expectsInstalledHooks`.
- `Codable` on `CodexHookFileMutation` / `HookHealthReport` per CLAUDE.md is a pre-existing gap; fine to leave for a separate PR.

PR #645, opened 2026-08-06
fix: distinguish Cursor from VS Code and fail closed on unknown hosts

KesleyDavid:
## Summary

Fixes #511 (multiple IDE jump / Cursor mislabeled as VS Code).

### Root causes

1. **Cursor → VS Code mislabel:** Cursor’s integrated terminal sets `TERM_PROGRAM=vscode`. Claude hooks already checked `CURSOR_TRACE_ID`, but **Codex** and **Gemini** always returned `"VS Code"`. Sessions then jumped with the VS Code bundle/CLI instead of Cursor.
2. **Wrong terminal on unmatched host:** `TerminalJumpService.resolveTerminalApp` fell through to the **first installed** known app when the reported name didn’t match, so multi-terminal setups often “always switched to one terminal”.

### Changes

- New `TerminalAppInference` helper (shared Cursor signals: `CURSOR_*`, Cursor bundle id, `Cursor.app` in VS Code helper paths).
- Claude / Codex / Gemini hooks use the shared VS Code-family disambiguation.
- `resolveTerminalApp` fails closed (returns `nil`) for unmatched names → Finder/cwd fallback instead of a random terminal.
- Tests for inference + unmatched host jump behavior.

### Verification

```bash
swift test --filter 'TerminalAppInferenceTests|TerminalJumpServiceTests'
```

All targeted tests passed (Ghostty live integration remains skipped without env flag).

Drafted with AI assistance; reviewed by me (KesleyDavid).

## Summary by CodeRabbit

* **New Features**
  * Added support for detecting Zed as a terminal host.
  * Improved identification of VS Code-family terminals, including Cursor, using available environment and application signals.

* **Bug Fixes**
  * Unrecognized terminal names now open the working directory in Finder instead of selecting an unrelated installed terminal.
  * Preserved accurate classification for standard VS Code environments.
  * Improved error details when terminal automation fails.

PR #674, opened 2026-09-01
fix: resolve VS Code forks so jump-back targets the real editor

danielhabila:
## Problem

Clicking a session row does nothing for Claude Code sessions run inside Cursor.

Every VS Code fork inherits `TERM_PROGRAM=vscode`, and `inferTerminalApp` only separated Cursor from upstream VS Code via `CURSOR_TRACE_ID`. Cursor 3.18.9 doesn't export that variable, so the session is tagged `VS Code`, the jump resolves `com.microsoft.VSCode`, and on a machine with no upstream VS Code installed every branch fails:

- `jumpToVSCodeFamilyWorkspace` runs `code -r <dir>` → no such CLI
- `appIsRunning` is false → no activation
- the working-directory fallback runs `open -b com.microsoft.VSCode <dir>` → throws

The user sees a row that swallows clicks.

## Changes

**`VSCodeFamilyHost`** (new, `OpenIslandCore`) narrows an already-identified VS Code family host to the actual fork, using `__CFBundleIdentifier` and then the bundled git askpass paths (`VSCODE_GIT_ASKPASS_NODE` / `GIT_ASKPASS`, which point into the hosting app bundle). Wired into the Claude, Codex, and Gemini hook detectors, all three of which had the same `case "vscode": return "VS Code"`.

Those signals leak across apps through macOS GUI environment inheritance, which is exactly what the existing `TERM_PROGRAM` comments warn about — so they are consulted *only* after `TERM_PROGRAM` has already placed the host in the VS Code family. They pick a family member; they never pick the family.

**Bundled CLI resolution** in `TerminalJumpService`: `jumpToVSCodeFamilyWorkspace` now prefers `<App>.app/Contents/Resources/app/bin/<cli>` before falling back to the bare name. A GUI-launched app inherits the LaunchServices `PATH` (`/usr/bin:/bin:/usr/sbin:/sbin`), which contains neither `/usr/local/bin` nor Homebrew, so `cursor -r <dir>` could not be resolved even with the right editor identified — the jump silently degraded to plain app activation and lost the workspace targeting.

## Verification

- `swift test` — new `VSCodeFamilyHostTests`, two `ClaudeHooksTests` regressions (Cursor without the trace marker; upstream VS Code still resolves to VS Code), and a `TerminalJumpServiceTests` case asserting the in-bundle launcher is preferred with the right arguments. Full suite (376 tests) passes.
- `zsh scripts/harness.sh lint docs` passes.
- The environment in the new Cursor fixtures was captured from a real Cursor 3.18.9 integrated-terminal shell on macOS 26.2: `TERM_PROGRAM=vscode`, no `CURSOR_TRACE_ID`, `__CFBundleIdentifier=com.todesktop.230313mzl4w4u92`, `GIT_ASKPASS=/Applications/Cursor.app/...`. That machine has no upstream VS Code installed, which is where the jump failure was originally observed.
- Not yet exercised end-to-end on the other two hook sources (Codex, Gemini); the change there is the same one-line call.

Not fixed here: VS Code family editors expose no way to focus a specific integrated terminal from outside, so the jump lands on the workspace window, not the exact terminal tab.

## Summary by CodeRabbit

* **New Features**
  * Improved detection of VS Code-family editors, including Cursor, Windsurf, and other compatible forks.
  * Terminal sessions now display the specific editor hosting them instead of a generic VS Code label.
  * Workspace jumping now prefers the editor’s bundled command-line launcher for more reliable navigation.

* **Bug Fixes**
  * Improved workspace focus behavior when launched from graphical applications with limited PATH settings.

### DIFF
--- a/Sources/OpenIslandApp/TerminalJumpService.swift
+++ b/Sources/OpenIslandApp/TerminalJumpService.swift
@@ -477,7 +477,37 @@ struct TerminalJumpService {
         guard let cli = Self.vscodeFamilyCLI[bundleIdentifier] else {
             return false
         }
+
+        // Prefer the launcher shipped inside the app bundle. A GUI-launched app
+        // inherits the LaunchServices PATH (/usr/bin:/bin:/usr/sbin:/sbin),
+        // which contains neither /usr/local/bin nor Homebrew, so resolving the
+        // CLI by bare name fails for most installs and the jump silently
+        // degrades to plain app activation.
+        if let bundledCLI = bundledVSCodeFamilyCLIPath(for: bundleIdentifier, cli: cli),
+           processRunner(bundledCLI, ["-r", workspacePath]) {
+            return true
+        }
+
         return processRunner(cli, ["-r", workspacePath])
+    }
+
+    /// Path to a VS Code family editor's `Contents/Resources/app/bin/<cli>`
+    /// launcher, when the app is installed and ships one.
+    private func bundledVSCodeFamilyCLIPath(for bundleIdentifier: String, cli: String) -> String? {
+        guard let appURL = applicationResolver(bundleIdentifier) else {
+            return nil
+        }
+
+        let cliPath = appURL
+            .appendingPathComponent("Contents/Resources/app/bin")
+            .appendingPathComponent(cli)
+            .path
+
+        guard FileManager.default.isExecutableFile(atPath: cliPath) else {
+            return nil
+        }
+
+        return cliPath
     }
 
     // MARK: - JetBrains IDE family
--- a/Sources/OpenIslandCore/ClaudeHooks.swift
+++ b/Sources/OpenIslandCore/ClaudeHooks.swift
@@ -1220,14 +1220,14 @@ public extension ClaudeHookPayload {
             case "wezterm":
                 return "WezTerm"
             case "vscode":
-                // Cursor also sets TERM_PROGRAM=vscode; check its unique
-                // env var first.
-                if environment["CURSOR_TRACE_ID"] != nil {
-                    return "Cursor"
-                }
-                return "VS Code"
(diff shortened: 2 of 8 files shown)


Octane0411:
Thanks — of the two VS Code-family PRs this is the better design (#645 overlaps; see my note there): table-driven, covers Cursor/Windsurf/Trae/Insiders, and fixes the real GUI-PATH failure by launching the CLI bundled inside the editor app. Built and ran the suite locally on top of current main: 377 tests pass.

**Needs a change before merge**
- `VSCodeFamilyHost` can now return `"Qoder"`, but `TerminalJumpService.knownApps` has no Qoder descriptor and `ProcessMonitoringCoordinator.supportedTerminalApp(for:)` has no `qoder` case. A Qoder-hosted session therefore falls through `resolveTerminalApp` to the first installed known app (iTerm/Ghostty/Terminal) — a regression from the previous "VS Code" label, which at least targeted a VS Code-family editor. Either add `TerminalAppDescriptor("Qoder", "com.qoder.qoder", ["qoder"])` + the `supportedTerminalApp` case + README rows (README currently lists Qoder only as an agent, not as an IDE host), or drop Qoder from the resolver for now.

**Optional**
- `VSCodeFamilyHost.swift:52-60` checks `CURSOR_TRACE_ID` before `__CFBundleIdentifier` / askpass paths. Running `code .` from a Cursor terminal leaks `CURSOR_TRACE_ID` into VS Code, whose shells carry `__CFBundleIdentifier=com.microsoft.VSCode` yet would be labelled Cursor. Bundle id / askpass are per-process and should outrank the marker.
- A Codex/Gemini regression test for "Cursor without trace id" would be nice; only Claude has one today.

Once the Qoder point is addressed this is good to merge.

PR #681, opened 2026-09-02
feat: opt-out compact usage display, keep detailed as default

Kevindic0214:
## Problem

#675 replaced the opened island's peak-only usage chip with one that always shows every window (5h + 7d, with reset countdowns). There's no way back to a lighter, peak-only chip for people who preferred the smaller footprint.

## Change

Follow-up suggested on #635 / #636: split `IslandUsageDisplay` into `.compact` and `.detailed`.

- `.detailed` is the new default and keeps #675's behavior exactly: the full `ViewThatFits` ladder (all windows + reset countdown, degrading to short titles, then to peak-only as a last-resort space fallback).
- `.compact` is the new opt-out: it skips straight to the peak-only rungs (full title / short title, single window, no countdown) — the pre-#675 chip.
- `openedUsageProviders` now only excludes usage when `.hidden`, not specifically `.compact`.
- Settings preview updated to match: `.compact` still shows one window per provider (unchanged), `.detailed` shows one provider with both windows, each colored independently by its own `usedPercentage` threshold.

`#675` hasn't shipped in a release yet (main is past v1.1.8), so there's no persisted-preference migration concern — flipping the default to `.detailed` costs existing installs nothing since they haven't seen the old behavior in a release.

## Verification

- `swift build` — passes.
- `scripts/lint-strings.sh`, `scripts/check-docs.sh` — pass.
- `swift test` — could not run locally (no `swift-testing` module on this toolchain; same environment gap noted on #635/#636, reproduces on a pristine main). CI covers the test run.
- Manual — harness smoke (`sessionList` scenario) with a temporary fake usage snapshot, confirming:
  - `.compact` → `Claude 7d 84%` (peak window only, no countdown)
  - `.detailed` → `Cl 5h 42% 7d 84%` (both windows, colored independently, short title kicking in at the same narrow width as before)

## Summary by CodeRabbit

* **New Features**
  * Added a **Detailed** usage display option for the expanded Island.
  * Detailed mode shows additional usage windows, countdowns, and provider information.
  * Usage information is available in Compact and Detailed modes; Hidden mode continues to conceal it.
  * The default usage display is now Detailed.

* **Documentation**
  * Updated appearance setting descriptions and translations to explain the available usage display levels.

### DIFF
--- a/Sources/OpenIslandApp/AppModel.swift
+++ b/Sources/OpenIslandApp/AppModel.swift
@@ -567,7 +567,7 @@ final class AppModel {
             usageDisplay: IslandUsageDisplay(
                 rawValue: defaults.string(forKey: appearanceDefaultsKey(profile, "usageDisplay"))
                     ?? ""
-            ) ?? .compact,
+            ) ?? .detailed,
             sessionStateIndicator: IslandSessionStateIndicator(
                 rawValue: defaults.string(forKey: appearanceDefaultsKey(profile, "stateIndicator"))
                     ?? defaults.string(forKey: legacyIslandSessionStateIndicatorDefaultsKey)
--- a/Sources/OpenIslandApp/AppModelTypes.swift
+++ b/Sources/OpenIslandApp/AppModelTypes.swift
@@ -56,7 +56,7 @@ enum IslandAppearanceDisplayProfile: String, CaseIterable, Identifiable, Sendabl
 struct IslandAppearancePreferences: Equatable, Sendable {
     var rightSlot: IslandRightSlot = .count
     var centerLabel: IslandCenterLabel = .agentAction
-    var usageDisplay: IslandUsageDisplay = .compact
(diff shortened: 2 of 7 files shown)


Kevindic0214:
Both CodeRabbit findings were valid — fixed in 69fff5c.

- **`loadAppearancePreferences` fallback** (major): real bug. The declared `.detailed` default never reached runtime because the UserDefaults load fell back to `.compact` when no key is stored, so a fresh profile would have loaded compact mode. Verified by deleting the stored key and running the harness: the chip now renders `Cl 5h 42% 7d 84%` instead of the peak-only `Claude 7d 84%`.
- **Ladder comment** (minor): the doc claimed countdowns drop before the title abbreviates; the rungs actually shorten the title first (rung 2 keeps `showsRemaining: true`). Reworded to match — no behavior change. Note this inaccuracy predates the PR; it came in with #675 and I carried it forward when rewording.

PR #682, opened 2026-09-02
fix: stop reposting outside clicks that already reached other apps

hnrobert:
## Problem

When a notification opens the island, the user's next click anywhere else on screen "jumps" — it lands as if it were a double click. Typical symptoms: clicking once in a terminal or editor selects a whole word, buttons fire twice. It was easy to mistake this for macOS forcing a focus switch when the notification appears.

## Root cause

There is **no** focus switch on the notification path — the panel is a `.nonactivatingPanel` presented with `orderFrontRegardless()`, and `NSApp.activate` only exists in the settings-window path.

The duplicate click came from our own event reposting:

1. A notification opens the island (panel interactive, app stays in the background — correct so far).
2. The user clicks somewhere outside the island. That click is delivered **normally** to the app under the cursor — `NSEvent.addGlobalMonitorForEvents` can only observe events, never consume them.
3. But the global click monitor handler also ran `handleMouseDown`, which called `notchClose()` **and unconditionally `repostMouseDown(at:)`**, injecting a second synthetic `leftMouseDown`/`leftMouseUp` pair via `CGEvent` at the same screen coordinates.
4. The target app receives two full click cycles ~10–30 ms apart → parsed as a double click → the "jump".

The repost is only legitimate when the original click never reached its target — i.e. when it was swallowed by our own panel window (local monitor case).

## Fix

`Sources/OpenIslandApp/OverlayPanelController.swift` (only file changed):

- `NotchEventMonitors` now reports where each click came from: the local monitor (event delivered to this app) passes `isLocalEvent: true`; the global monitor (event delivered to another app) passes `false`.
- `handleMouseDown(_:isLocalEvent:)` reposts **only when `isLocalEvent == true` and the click point lies inside the panel frame** — that is, only when the panel actually consumed the original click and nothing under the cursor received it.
- Doc comments on `handleMouseDown` and `repostMouseDown` now state the precondition explicitly.

This preserves the original purpose of the repost (restoring a click swallowed by the panel window) while eliminating the duplicate in the common notification case.

| Scenario | Before | After |
| --- | --- | --- |
| Island open (notification), click in another app | click **+ duplicate synthetic click** (double-click artifact) | single click, island collapses |
| Click swallowed by the panel frame | synthetic repost restores it | unchanged (still reposted) |

## Verification

- `swift build` — passes.
- `swift test` — 393 tests in 42 suites, all pass.
- Manual, via the real data path: launched the dev app from this branch, injected a synthetic `PermissionRequest` through `OpenIslandHooks --source claude` (Unix-socket bridge), then single-clicked in a text area outside the island — the island collapsed and the click registered exactly once (no word selection).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved overlay dismissal handling for mouse clicks.
  * Prevented duplicate clicks from being sent to other applications when interacting with the overlay.
  * Preserved clicks inside the overlay by replaying them only when necessary.
  * Improved click behavior when switching between the overlay and other applications.
  * Ensured clicks outside the overlay continue to reach the intended application without unintended duplication.

### DIFF
--- a/Sources/OpenIslandApp/OverlayPanelController.swift
+++ b/Sources/OpenIslandApp/OverlayPanelController.swift
@@ -232,8 +232,8 @@ final class OverlayPanelController {
 
         eventMonitors.start { [weak self] location in
             self?.handleMouseMoved(location)
-        } mouseDownHandler: { [weak self] location in
-            self?.handleMouseDown(location)
+        } mouseDownHandler: { [weak self] location, isLocalEvent in
+            self?.handleMouseDown(location, isLocalEvent: isLocalEvent)
         }
     }
 
@@ -261,7 +261,12 @@ final class OverlayPanelController {
         }
     }
 
-    private func handleMouseDown(_ screenLocation: NSPoint) {
+    /// - Parameter isLocalEvent: Whether the click was delivered to this app
+    ///   (local monitor) or to another app (global monitor). Global monitors
+    ///   only observe — the clicked app already received the event — so the
+    ///   synthetic repost must be skipped there or it lands as a duplicate
+    ///   click (double-click word selection, double activation).
+    private func handleMouseDown(_ screenLocation: NSPoint, isLocalEvent: Bool) {
         guard let model else { return }
 
         let inClosedSurfaceArea = isPointInClosedSurfaceArea(screenLocation)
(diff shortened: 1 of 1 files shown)


hnrobert:
@Octane0411 👀

This improves my experience quite a lot locally, tested, worth trying.

Shall I do a rebasing here?

hnrobert:
🧐

hnrobert:
Any suggestion on this? @Octane0411 👀

Issue #691, opened 2026-09-04
feat: Support QoderCN (Chinese distribution of Qoder)


jerryyon-eng:
## Feature Description

QoderCN (the China distribution of Qoder, qoder.cn) ships as a separate app bundle with its own bundle identifier and config directory. Open Island currently recognizes only the international Qoder, so QoderCN sessions never appear on the island. Requesting first-class support for QoderCN, analogous to the existing Qoder integration.

## Environment (verified on my machine)

- Open Island 1.2.0 (build 80), macOS 26.5, Apple Silicon
- QoderCN 0.1.6 at `/Applications/Qoder CN.app`, bundle ID `com.qodercn.app`
- Hooks were installed into `~/.qoder/settings.json` (international path) by the Settings window, while `~/.qoder-cn/settings.json` (the one QoderCN actually reads) contains no hooks

## Motivation

I'm a QoderCN user and the island shows nothing for my sessions. I diagnosed three integration points that hardcode the international Qoder identity, and all of them fail to match QoderCN:

| # | Integration point | Expects (international Qoder) | QoderCN actual |
|---|---|---|---|
| 1 | Hook config path | `~/.qoder/settings.json` | `~/.qoder-cn/settings.json` |
| 2 | Agent process discovery (`ActiveAgentProcessDiscovery.swift:565`) | process path contains `/qoder.app/` | app is `Qoder CN.app` → `/qoder cn.app/` (note the space) |
| 3 | Jump-back bundle-ID map (`TerminalJumpService.swift:482`) | `com.qoder.qoder` | `com.qodercn.app` |

Since #1 never fires, no hook events reach the island at all — #2 and #3 never even get the chance.

**Related bug (international Qoder)**: while verifying, I found the real bundle ID of international Qoder 0.1.6 is `com.qoder.app` (read from its `Info.plist`), but `vscodeFamilyCLI` maps `com.qoder.qoder`. Unless older Qoder builds used that identifier, jump-back currently never matches even for international Qoder users. This might deserve a separate `fix:` PR — happy to send one.

## Proposed Behavior

1. **Hooks**: add a `qodercn` source writing to `~/.qoder-cn/settings.json` (same `OpenIslandHooks` binary); payload decode can reuse the Claude-family path exactly like `qoder` does
2. **Process discovery**: match `/qoder cn.app/` alongside `/qoder.app/`
[...]

PR #692, opened 2026-09-04
fix(qoder): map real bundle ID (com.qoder.app) for jump-back

jerryyon-eng:
## Summary

- `vscodeFamilyCLI` maps the VS Code-family jump-back CLI by bundle identifier, but Qoder's only key was `com.qoder.qoder`. Qoder 0.1.6 (current release) actually reports `com.qoder.app` (read from `/Applications/Qoder.app/Contents/Info.plist`), so the workspace jump-back path never matched and fell through to other handling.
- Adds `com.qoder.app → qoder` alongside the existing key. The old key is kept in case earlier Qoder builds shipped that identifier.

Found while diagnosing #691 (QoderCN support request). QoderCN's separate bundle ID (`com.qodercn.app`) is part of the same map, but that likely deserves its own PR after maintainer input on the jump approach for a CLI-less app.

## Verification

- `swift build` → Build complete
- `swift test` → 447 tests in 48 suites, all passed (macOS 26.5, Xcode 26.6, Swift 6.3.2)

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing Qoder as a VS Code-family terminal application.
  * Workspace jumps can now open and activate Qoder when it is the selected application.

### DIFF
--- a/Sources/OpenIslandApp/TerminalJumpService.swift
+++ b/Sources/OpenIslandApp/TerminalJumpService.swift
@@ -121,6 +121,12 @@ struct TerminalJumpService {
             ]
         ),
         TerminalAppDescriptor(
+            displayName: "Qoder",
+            bundleIdentifier: "com.qoder.app",
+            aliases: ["qoder"],
+            alternateBundleIdentifiers: ["com.qoder.qoder"]
+        ),
+        TerminalAppDescriptor(
             displayName: "Zed",
             bundleIdentifier: "dev.zed.Zed",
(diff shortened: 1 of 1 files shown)


jerryyon-eng:
Addressed the CodeRabbit finding: added a Qoder descriptor to `knownApps` (primary `com.qoder.app`, alternate `com.qoder.qoder`) so Qoder jump targets actually reach the VS Code-family dispatch. `swift build` + `swift test` (447 tests) pass.

PR #694, opened 2026-09-04
feat(qoder-cn): add QoderCN (Chinese distribution) adapter

jerryyon-eng:
Implements #691 — first-class support for QoderCN, the China distribution of Qoder (separate app bundle `Qoder CN.app`, bundle ID `com.qodercn.app`, config directory `~/.qoder-cn/`).

## What's included

- **Agent identity**: new `.qoderCN` case across `AgentIdentifier` / `AgentTool` (display name "QoderCN", short name "QODERCN", brand color `#ff4d6d` — kin to Qoder's `#ff6b9f`), `isClaudeCodeFork`, permission summaries, session naming and completion-recipient coverage.
- **Hooks**: hooks CLI accepts `--source qodercn` (Claude-format payload decode, same as `qoder`); `HookInstallationCoordinator` gains a manager targeting `~/.qoder-cn/settings.json` with full install / uninstall / status / auto-install wiring; Settings gains a QoderCN row (with uninstall confirmation) and install-all coverage.
- **Process discovery**: recognizes processes under `/qoder cn.app/`.
- **Jump-back**: `com.qodercn.app` registered in `knownApps` and handled with app-level activation (`open -b`), since QoderCN ships no CLI launcher — mirroring the Conductor approach proposed in #691. Intentionally **no** `vscodeFamilyCLI` entry, so this PR stays independent of #692.

## Verification

- `swift build` clean
- `swift test` → 447 tests in 48 suites, all passed (macOS 26.5, Xcode 26.6, Swift 6.3.2)
- `AgentSessionPresentationTests` exhaustive-coverage test extended for `.qoderCN`
- Hook display path for QoderCN was already manually verified on this machine (hooks written by hand before this adapter — sessions appear on the island with live status); this PR adds the identity, discovery, jump-back and Settings management around it.

Happy to adjust naming (e.g. `qodercn` source string, "QoderCN" display name) or the jump-back approach per maintainer preference.

## Summary by CodeRabbit

* **New Features**
  * Added support for the QoderCN agent across session detection, status displays, permissions, and approval workflows.
  * Added QoderCN hook installation, uninstallation, status tracking, automatic setup, and migration support.
  * Added QoderCN controls in Settings, including “Install All” support.
  * Added QoderCN app activation and terminal navigation support.
  * Improved terminal selection for fallback navigation.
* **Tests**
  * Expanded coverage for QoderCN detection and agent handling.

### DIFF
--- a/Sources/OpenIslandApp/ActiveAgentProcessDiscovery.swift
+++ b/Sources/OpenIslandApp/ActiveAgentProcessDiscovery.swift
@@ -565,6 +565,9 @@ struct ActiveAgentProcessDiscovery {
         if lowered.contains("/qoder.app/") {
             return "Qoder"
         }
+        if lowered.contains("/qoder cn.app/") {
+            return "QoderCN"
+        }
         if lowered.contains("/codebuddy.app/") {
             return "CodeBuddy"
         }
--- a/Sources/OpenIslandApp/AgentSession+Presentation.swift
+++ b/Sources/OpenIslandApp/AgentSession+Presentation.swift
@@ -218,6 +218,8 @@ extension AgentSession {
             return "OpenCode"
         case .qoder:
             return "Qoder"
+        case .qoderCN:
+            return "QoderCN"
         case .qwenCode:
             return "Qwen Code"
         case .factory:
--- a/Sources/OpenIslandApp/AppModel.swift
+++ b/Sources/OpenIslandApp/AppModel.swift
@@ -98,14 +98,17 @@ final class AppModel {
     var codexHooksInstalled: Bool { hooks.codexHooksInstalled }
     var claudeHooksInstalled: Bool { hooks.claudeHooksInstalled }
     var qoderHooksInstalled: Bool { hooks.qoderHooksInstalled }
+    var qoderCNHooksInstalled: Bool { hooks.qoderCNHooksInstalled }
     var qwenCodeHooksInstalled: Bool { hooks.qwenCodeHooksInstalled }
     var factoryHooksInstalled: Bool { hooks.factoryHooksInstalled }
     var codebuddyHooksInstalled: Bool { hooks.codebuddyHooksInstalled }
     var qoderHookStatus: ClaudeHookInstallationStatus? { hooks.qoderHookStatus }
+    var qoderCNHookStatus: ClaudeHookInstallationStatus? { hooks.qoderCNHookStatus }
     var qwenCodeHookStatus: ClaudeHookInstallationStatus? { hooks.qwenCodeHookStatus }
     var factoryHookStatus: ClaudeHookInstallationStatus? { hooks.factoryHookStatus }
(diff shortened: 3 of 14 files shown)


jerryyon-eng:
Addressed all three findings:

1. **Normalization** — added `qoder` / `qodercn` / `qoder cn` cases to `supportedTerminalApp` (the international `qoder` label was also missing there — it degraded to "Unknown" too, which is what the manually-installed-hooks session showed). Regression test added.

2. **zellijParentTerminals** — added an `isTerminalHost` flag to `TerminalAppDescriptor` (default `true`) and excluded Codex.app / Claude.app / Conductor from the parent-terminal fallback. **Partial divergence on QoderCN itself**: like the VS Code-family entries already in the list, QoderCN is an Electron IDE with an integrated terminal, so it can legitimately host a zellij session — I kept it as a terminal host for consistency with Cursor/VS Code. Happy to flip it if you'd rather treat IDEs as non-hosts.

3. **Titles** — `sessionTitle`, `permissionRequestTitle` (both Claude-branded cases) and the multi-question prompt now use QoderCN labels when the source is `qodercn`; every other source keeps its exact existing strings.

`swift test`: 448 tests in 48 suites, all passed.

jerryyon-eng:
Addressed the third-round finding — with one correction to the suggested fix: the descriptor should use `com.qoder.app` as primary (the real bundle ID verified from Qoder 0.1.6's Info.plist), with `com.qoder.qoder` as an alternate, matching #692 exactly. A `com.qoder.qoder`-only descriptor would reintroduce the wrong-ID bug #692 fixes.

Verified the fallback hazard first: `resolveTerminalApp` does fall back to `knownApps.first(where: isInstalled(descriptor:))` when the exact name misses, so without a descriptor the new "Qoder" label could activate an unrelated running app. The descriptor is therefore necessary in this PR even though it overlaps #692 — whichever PR merges second drops the duplicate during rebase.

`swift test`: 448 tests in 48 suites, all passed.

jerryyon-eng:
Addressed the fourth-round finding, both parts:

1. **vscodeFamilyCLI** — added `com.qoder.app → qoder` alongside the legacy key. Verified: with the descriptor from the previous commit resolving to the primary ID, the missing routing entry meant primary installs skipped `jumpToVSCodeFamilyWorkspace` entirely.
2. **zellijParentTerminals** — now derived from `allBundleIdentifiers` (flatMap) so alternate installs (Trae CN, legacy Qoder, Zed Preview) are covered by the fallback.

Note on overlap: with these two commits, this PR now fully contains the changes from #692 (corrected Qoder descriptor + both bundle IDs in `vscodeFamilyCLI`). That overlap became necessary once the descriptor landed here — without the routing entry, primary Qoder installs had no workspace-jump path. If you'd rather merge just this PR, closing #692 is fine — its findings are all credited here and in the linked issue.

`swift test`: 448 tests in 48 suites, all passed.

PR #710, opened 2026-09-17
feat: add ZCode hook integration (14th agent)

zephyrq-z:
## Summary

Adds **ZCode** (ZCode Desktop, `ZCode.app`) as the 14th supported agent.

ZCode's hook payloads are **Claude Code compatible on stdin** — snake_case `hook_event_name` / `session_id` / `cwd`, tool events carry `tool_name` / `tool_input`, `Stop` carries `last_assistant_message` (verified against the engine source shipped inside `ZCode.app/Contents/Resources/glm/zcode.cjs`). So the runtime reuses the existing `ClaudeHookPayload` decode path behind a new `--source zcode` value — no new decoder, no `TerminalJumpService` changes.

What differs is the **configuration file**: ZCode reads hooks from the top-level `hooks` block of `~/.zcode/cli/config.json`, shaped `{ enabled?, events: { <Event>: [group] } }`, and configuration-file hooks are **disabled by default**. This PR adds a dedicated installer for that shape.

## What's included

- **`ZCodeHookInstaller`** — managed install of the low-noise lifecycle event set (`SessionStart`, `UserPromptSubmit`, `Stop`, same set as Codex), sets `hooks.enabled: true`, preserves user-authored hook groups in the same file, uninstalls by exact manifest command + legacy-command match, and drops `enabled` when nothing else remains
- **`ZCodeHookInstallationManager`** — manifest, timestamped backup, managed-binary install (same pattern as Kimi/Grok managers)
- `AgentTool` / `AgentIdentifier`: `.zcode` case with display name, short name, brand color
- `ClaudeHooks.resolvedAgentTool`: `--source zcode` → `.zcode`
- `HookInstallationCoordinator` / `SettingsView` / `AppModel`: status, install/uninstall, intent-store auto-repair wiring
- `ActiveAgentProcessDiscovery`: matches the `ZCode.app` Electron process and its embedded engine (`…/Resources/glm/zcode.cjs`); liveness keeps tracked zcode sessions alive while the app runs (same conservative fallback as Kimi, same app-process shape as Codex.app)
- `OpenIslandSetup`: `installZcode` / `statusZcode` / `uninstallZcode` with `--zcode-dir`
- Docs: README (en + zh-CN), `docs/hooks.md` (new ZCode section + timeout table row)

## Scope notes (deliberate v1 boundaries)

- **No `PreToolUse` / `PermissionRequest` registration** — ZCode supports blocking events, but approval round-trips are not designed for this surface yet. Keeping the registered set lifecycle-only matches the Codex low-noise pattern and keeps this slice small.
- **No usage tracking** — ZCode does not expose a Claude-style usage file.
- `transcript_path` in ZCode payloads points at a hook-runtime temporary file deleted after the hook completes; nothing depends on it.

## Verification

- New tests: installer round-trip (JSON shape, user-entry preservation, idempotent install, legacy command recognition), payload decode, `--source zcode` → `.zcode` mapping — 11/11 pass
- Full suite: **462/462 pass**, including the `completionReplyRecipientCoversEveryAgentTool` coverage guard
- Setup CLI verified end-to-end in a sandbox dir (install → status → uninstall restores the original config exactly)
- **Real-home verification**: installed into `~/.zcode/cli/config.json`, launched the dev app, and confirmed via a bridge observer client that fired hook events produce the full broadcast stream — `sessionStarted(tool: "zcode")`, `activityUpdated("Prompt: …")`, `sessionCompleted(last_assistant_message)`; also confirmed fail-open (hook exits 0 with the bridge down) and `agentIntent.zcode = installed` recording on app startup

Tested against ZCode.app `glm/zcode.cjs` hook runner on macOS 15.

## Summary by CodeRabbit

- **New Features**
  - Added ZCode and Hermes Agent support for session tracking and hook-based integrations.
  - Added Settings controls to install, check, and uninstall hooks for both agents.
  - Added CLI commands for hook management, including custom directory support.
  - Added automatic hook setup and Hermes process/session monitoring.
  - Added Hermes metadata, completion handling, questions, and permission prompts.
- **Documentation**
  - Updated English and Chinese README files and hook documentation with compatibility and setup instructions.


zephyrq-z:
Verified this branch against the Hermes Agent integration from #706 (`dragonkid:feat/hermes-agent-integration` @ e5eab29) by merging it locally.

**Conflict resolution** — all 14 conflicted files were parallel-insert conflicts (both PRs adding a 14th agent); resolved by keeping both sides, zcode before hermes in every enum/wiring ordering. README agent counts updated to **15 agents**.

**Merged-tree verification:**
- `swift build` clean, full suite **517/517 tests pass** (462 zcode-side + 55 Hermes tests, zero regressions)
- `OpenIslandSetup statusZcode` / `statusHermes` both report correctly on the merged build
- Re-ran the live bridge-observer check against the merged dev app: zcode SessionStart/Stop events still produce the full `sessionStarted` → `sessionCompleted` broadcast stream

No cross-interference found between the two integrations — they touch disjoint decoders (`ClaudeHookPayload` vs `HermesHookPayload`) and disjoint config files. Reviewers can treat the two PRs as independent; this merge is shared for anyone who wants to run both locally.

zephyrq-z:
Addressed all 6 CodeRabbit findings (5ae800c):

1. **TTY-less ZCode processes (Major)** — added `!isZcodeProcess(...)` to the discovery TTY guard, alongside the existing OpenCode/Hermes exemptions. Without this, the Electron process (never has a TTY) was filtered out before detection and active zcode sessions could be evicted while ZCode.app runs.
2. **Exact managed-hook removal signal (Major)** — uninstall now detects removal at the hook level (catches a managed hook sharing a group with user hooks, which group-count comparison missed), and `Manager.status` reports `managedHooksPresent` instead of `changed` so a reformatted-but-unmanaged config is no longer mistaken for an install.
3. **`ZCodeHookFileMutation` Codable (Major)** — added per the repository model rules.
4. **Agent-labeled session titles (Minor)** — `sessionTitle` and the fallback `paneTitle` now use `resolvedAgentTool.displayName`; verified live via the bridge observer: sessions now start as `ZCode · <workspace>` instead of `Claude · <workspace>`.
5. **Resolved binary URL in setup output (Minor)** — `installZcode`/`statusZcode` print `status.hooksBinaryURL` (the managed path), not the source path.
6. **Docs alignment (Minor)** — hooks.md now says ZCode registers the Codex lifecycle set *minus* `PermissionRequest`, since the Codex managed install includes it.

Added two tests for #2 (shared-group removal, exact-signal status). Full suite: **519/519**.

